PR #86, opened 2013-11-26
fully fledged selection capabilities

bolshchikov:
This is a mixin for a component which provides complete functionality for the selection behavior:

* Support of `ctrl` and `cmd` key multiple selection
* Support of `shift` key
* Support of `ctrl+a` / `cmd+a` combination to select all records
* Support of arrows `up` and `down` to switch selection from the keyboard
* ContextMenu event handling and selection:
    1. If click is on the row that is not selected, the clicked row becomes the selection
    2. If click is on the row that currently is in the list of selection, selection does not change

The component also accepts action `contextMenuHandler` which will be invoked when contextMenu event occurs, e.g.

    {{table component contextMenuHandler="MyHostedControllerHandler"}}


azirbel:
I forgot to come back and update this PR. A while ago @sreenath-addepar took a lot of the ideas here and expanded them in https://github.com/Addepar/ember-table/pull/190, which has now been merged.

The goal of that #190 was to add some fundamental improvements to how selection works, while making sure to avoid bugs. It doesn't provide all the features sketched out here, but does provide multi-select. Ideas for using arrow keys are now being worked on at https://github.com/Addepar/ember-table-addon/pull/17, and there's some good progress on that.

@bolshchikov, I'm sorry we didn't end up merging this PR, but I wanted to thank you because I know a lot of your ideas were the foundation for https://github.com/Addepar/ember-table/pull/190.

Since the selection behavior has diverged a lot from here, it's time to close this PR. New selection features, as well as rebases of features from this PR onto the current architecture, are welcome.

PR #127, opened 2014-04-07
Process to build code examples

FellowMD:
NOTE: this is not ready to be merged. First off, I haven't finished converting all the examples, I just wanted to get the PR up for feedback.

A while back I noticed that one of the code examples was out of sync with what was actually running on the page (the application.hbs &lt;pre> tag wasn't what ember was actually rendering).

I put together a grunt task to pull the code in to both the displayed section and running area.

It's not too complicated. I looked around for a task that would do this but couldn't find one. The closest I found was usemin, which wasn't quite what I was looking for. 

The task looks for filepaths like `@@this/will/be/escaped.hbs@@`  or `@@@this/will/be/unescaped.hbs@@@` and replaces them with the referenced file.

What do you guys think? There are a few drawbacks and a bit of work left before I'd want you to merge, but it's a bit late tonight so I'll come back and list them tomorrow.


azirbel:
Thanks for this idea @FellowMD! I implemented it during the Ember CLI conversion, here: https://github.com/Addepar/ember-table/pull/309.

PR #159, opened 2014-05-20
Fix for horizontal scroll

rinchik:
Horizontal scroll brakes if in last column canAutoResize is set to true.
Honoring minWidth will fix that.


azirbel:
Forgot to update this - I fixed the problem in https://github.com/Addepar/ember-table/commit/3dd725419338d22b75223d1a5fc9174cb0bb7542; `minWidth` should be honored now.

PR #173, opened 2014-06-18
Mentions supported browsers in readme

jrhe:
Specifies browsers we aim to support in readme.


azirbel:
Merged via https://github.com/Addepar/ember-table/commit/331951b405f616821c1ebcae6e2cc0d9d1bcb7e9.

Issue #184, opened 2014-07-17
Simple sort example


jasonf20:
Hello,

I want to create the simplest form of sort in my application. Clicking the header should sort by that column. 

However I can't seem to figure out how best to do this. 

I have made a solution that works, but I think its not how it's supposed to be done. 
(My solution extends the TableComponent to set the sort action - but I think this default behaviour should not require this)

The lack of documentation and examples of features makes it quite difficult.

Could an example of Simple sorting be added?

billy-addepar:
This version of Ember table is no longer supported. If you want to continue discussion, you can open the issue on [Ember Table Legacy](https://github.com/Addepar/ember-table-legacy)

PR #186, opened 2014-07-18
Added sorting support by clicking on table cell header

jasonf20:
As discussed in #184 I have implemented some simple sorting which can be merged in.

There is one thing I'm still not happy about (because I did not understand how to make it work):
I would like to have  default compare value method, something like: 
`return col.getCellContent(b) - col.getCellContent(a);`
but for whatever reason the content (a,b) are not ember objects and then they do not define the `.get()` method so the default implementation of  `getCellContent` fails.


gsasikumar:
Team, I know there has been lot going. We are also looking for the same functionality. In case we could help you test or improve we would be happy to do that. Please let us know.

billy-addepar:

PR #200, opened 2014-08-27
Resolve views that are referenced Ember CLI style.

gunn:
Ember table currently has a hardcoded requirement that view classes be attached to a global objects, and so it's currently not possible to use it in Ember CLI projects which avoid them.

This patch adds Ember CLI style resolving while supporting the current global one. All these are valid now:

``` javascript
tableCellViewClass: 'App.MessagePreviewView'  // Global Style
tableCellViewClass: 'message/preview'         // Ember CLI Style
tableCellViewClass: Ember.Table.TableCell.extend({ templateName: 'message/preview' })
```

Also, I've added alias' because I think it's more semantic to say:

``` javascript
tableCellView: 'message/preview'
```


glennfu:
Here's a bit more clarification for someone else trying to use this:

1. These classes (apparently) need to live within the /views folder. I couldn't get them to work anywhere else.
2. For tableRowView the Ember.Table.TableRow extension needs something like `classNameBindings: "row.rowClassName"` where your Model has rowClassName: "blah", defined as a string, not a function. 

Since I couldn't get any error messages when I tried other things, and couldn't find any documentation about it, I figured I wanted to add this in case someone else was trying to get this working.

PR #237, opened 2014-11-17
Update grunt-template-compiler to use Handlebars 2.0.0 and Ember.js 1.9.0-beta.3

intuitivepixel:


azirbel:
Thanks for once again leading the way @intuitivepixel.

I've set up a plan for supporting different Ember versions here: https://github.com/Addepar/ember-table/issues/288, and made an official 1.9 branch here: https://github.com/Addepar/ember-table/pull/286.

PR #242, opened 2014-12-30
Update contributing guidelines

Dylnuge:
@azirbel @igillis 

Add CoffeeScript compiling details to the contribution guidelines.

Let's critique this here; I'll open identical (s/table/charts and s/table/widgets) PRs for charts and widgets once we're entirely happy with the language.


azirbel:
LGTM! Are you suggesting we should trim down the file now? What you've added is great clarification.

Dylnuge:
I was just thinking that it might make the differences between our contrib policy and a "boilerplate" one more clear. It's a separate concern though, and the document is fine as is; I'll go ahead and merge this and make a simultaneous PR for charts/widgets

PR #245, opened 2015-01-06
Revert "Fix rendering issue in Linux Chrome v38"

azirbel:
This reverts commit ea0983f6707aec93eaa56482b037c64f748df578.

The fix caused a bug where header text would never wrap (long header text
should wrap to multiple lines).

@jiangty-addepar


jiangty-addepar:
Let's hope the Chrome team fixed this!

jiangty-addepar:
lgtm

PR #248, opened 2015-01-06
Change canAutoResize default to false.

azirbel:
@Dylnuge @igillis @cyril-sf 

I made a mistake with the new column changes and release 0.3.0. While `canAutoResize` has always defaulted to `true`, it used to be that you would activate automatic resizing by setting `forceFillColumns` to `true`. Now there is no such setting, because you can get all the same configuration directly from `canAutoResize` settings.

Defaulting to `false` is also the correct behavior, because it's the simplest way for a table to behave (columns are a fixed width and only change in width when you "tell them to" by dragging.


azirbel:
Note that I'll release a new version 0.4.0 once this change is merged in.

Dylnuge:
LGTM. Not 0.3.1?

azirbel:
I'm considering it a breaking change =/

PR #250, opened 2015-01-08
Re-fix rendering issue in Linux Chrome 38

jiangty-addepar:
The old fix had to be reverted (#245) because it introduced a bug. This fix doesn't re-introduce that bug.

Review: @azirbel


azirbel:
For context: even though this isn't the best fix, we decided it's ok to merge because it's unlikely to break other browsers, and fixes an issue due to a bug in Linux Chrome. It's hard to say whether a "good" fix exists when working around a browser bug, and this fix is well commented.

LGTM
Thanks @jiangty-addepar 

(That said, you forgot to add dist files, I'll just add them when I merge)

PR #251, opened 2015-01-08
Selection minor cleanup

azirbel:
Does two things:

* Use a simple array instead of ArrayProxy
* Don't add undefined elements to the selection
  * If `findRow` fails to find the row (e.g. because the `bodyContent` changed), it shouldn't add `undefined` to the `selection` array. It should at least leave the selection unchanged.

@ebryn @Dylnuge @cyril-sf


src/component.coffee
```diff
@@ -390,15 +392,13 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   # rows that were selected directly or as part of a previous
   # range selection (shift-click)
-  persistedSelection: Ember.computed ->
-    Ember.ArrayProxy.createWithMixins Ember.MutableArray, {content:[]}
+  persistedSelection: Ember.computed -> []
```
ebryn: You should use `Ember.A()` here if we care about this working for people with Ember's prototype extensions disabled

src/component.coffee
```diff
@@ -390,15 +392,13 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
 
   # rows that are part of the currently editable range selection
-  rangeSelection: Ember.computed ->
-    Ember.ArrayProxy.createWithMixins Ember.MutableArray, {content:[]}
+  rangeSelection: Ember.computed -> []
```
ebryn: Ditto above

azirbel:
@ebryn fixed

ebryn:
LGTM

src/component.coffee
igillis: One thing I'm going to do with ember-widgets is turn off prototype extensions for gh_pages so that people are forced to support it.

PR #254, opened 2015-01-09
Document new savedWidth and create migration guides

azirbel:
@Dylnuge @cyril-sf

### DIFF
--- a/app/templates/ember_table/migration-guides.hbs
+++ b/app/templates/ember_table/migration-guides.hbs
@@ -2,26 +2,42 @@
 <div class="col-md-10 col-md-offset-2 left-border main-content-container">
   <h1>Migration Guides</h1>
   <div class="row">
-    <div class="col-md-7">
+    <div class="col-md-9">
       <h2>Migrating from old versions to version 0.1.0</h2>
-      <p>Version 0.1.0 of Ember Table is a bit of a restructure - we’re hoping that the new organization will make it simpler and easier to set up, but for existing users, you’ll have to make a few changes to upgrade to the new Ember Table. This guide is intended to walk you through that process.</p>
-
+      <p>
+        Version 0.1.0 of Ember Table is a bit of a restructure - we’re hoping
+        that the new organization will make it simpler and easier to set up,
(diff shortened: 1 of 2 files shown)


Dylnuge:
I don't know how I feel about retroactively updating changelogs, because the changelog will still be incorrect on the `0.3.0` release/tag, but otherwise LGTM

azirbel:
On second thought, I share your concerns. I won't update the changelog for v0.3.0.

What I really want is a "migration guide" [here](http://addepar.github.io/ember-table/#/ember-table/migration-guides). I'll spend another few minutes and make that instead of messing with the changelog.

azirbel:
@Dylnuge @cyril-sf @igillis I tried to write a more serious migration guide instead. See what you think =)

Dylnuge:
LGTM.

app/templates/ember_table/migration-guides.hbs
```diff
@@ -2,26 +2,122 @@
...
+          defaults to <code>false</code>, but if a column sets it to
+          <code>true</code>, the force fill behavior is implicitly triggered.
+        </li>
+      </ol>
+
+      <h2>Migrating to version 0.3.0</h2>
+      <p>
+        Version 0.3.0 includes major column resizing changes, but with
+        <code>canAutoResize</code> defaulting to <code>true</code>.  We
+        recommend skipping this version and upgrading direcly to 0.4.0; see
```
Dylnuge: directly

cyril-sf:
Can you simply unregister the 0.3 release on a bower?

azirbel:
It was short lived, we release 0.4.0 the next day and put "skip 0.3" in both the changelog and migration guides. I don't think we need to unregister it (it's a working version), it just makes upgrading more annoying if you do it step by step and don't read the guides.

shaunc:
(IMHO unregistering w/ package managers is a bad idea. I recently got tied up in knots trying to install a package which included a sub-package w/ a version that had been deregistered in npm. The package owner couldn't see the problem because they already had installed locally.)

PR #259, opened 2015-01-19
Fix memory leaks

cyril-sf:
jQuery plugin event handlers need to be removed.

* with sortable
* with resizable

@addepar-andy @Addepar/uie @igillis @bjcohen


Dylnuge:
LGTM

Issue #243, opened 2015-01-05
Ember table fluid column reorder incompatibility with Ember 1.8.1


mariuserik:
Hi everyone.

After I've updated my application to ember 1.8.1, I got many problems when I reorder columns.

Always when I move a column to the first or to the last position, all my other columns disappear, and my table get messy.

I get the official example in jsbin from Addepar github page

and just changed this line:
``` <script src="http://builds.emberjs.com/tags/v1.4.0/ember.js"></script> ```
to this:
``` <script src="http://builds.emberjs.com/tags/v1.8.1/ember.js"></script> ```

And I get the same problem as in my application, as in screenshot:

![ember-table-problem](https://cloud.githubusercontent.com/assets/1163058/5613520/ae01496a-94e5-11e4-9f37-e2619d24cbb7.png)

I always get this problem with "previousSibling" or "nextSibling" depends if I move a column to the first place or the last 

This is the updated and problematic jsbin example:
http://emberjs.jsbin.com/misatufafe/1/edit?output

anyone have any solution/workaround to this problem?

Best Regards,

Marius

azirbel:
Hey, thanks so much for posting this! I don't know of a workaround , but I'll plan to take a look at it within the next couple days. I recently made some changes to column reordering, but it looks like the table breaks both with and without those changes.

mariuserik:
Hi @azirbel.

I appreciate your answer. And whatever i find something else on this topic i'll keep you updated.

mariuserik:
I've fixed this issue and sent a PR. 

https://github.com/Addepar/ember-table/pull/262/commits

azirbel:
Fixed by https://github.com/Addepar/ember-table/pull/265.

mariuserik:
@azirbel I've just removed my fix and tested the master branch, the reordering seems to work but I get some errors:
![reorder_columns](https://cloud.githubusercontent.com/assets/1163058/5899454/92aa384e-a55b-11e4-8cb9-7ea872201ac4.png)

Should I just ignore them?

azirbel:
Thanks for reporting, I'm seeing the problem too. I'm able to reproduce it by reordering the last column when the table is in fluid mode. I think this is happening because of some bug that miscalculates which columns are resizable. In fluid mode, the last column is not resizable because there's no column to the right it can steal width from. When you reorder it, it looks like the column is still not resizable, which is a bug and seems to be causing this error as well.

azirbel:
[...]

PR #262, opened 2015-01-21
Columns reorder fix, Ember 1.8.1 changed removeObject (Issue #243)

mariuserik:
Ember 1.8.1 is firing some methods from new Morph library when calling removeObject, and this was removing the columns from DOM, causing a null problem.

The problem is described here: https://github.com/Addepar/ember-table/issues/243

Using Jquery grep remove only the column to reorder, and prevent this from happening

### DIFF
--- a/src/component.coffee
+++ b/src/component.coffee
@@ -120,7 +120,10 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   onColumnSort: (column, newIndex) ->
     columns  = @get 'tableColumns'
-    columns.removeObject column
+
+    columns = $.grep columns, (n, i) ->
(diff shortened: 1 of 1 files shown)


azirbel:
Hi @mariuserik and thanks for bringing this to our attention (and for your fix)!

@ebryn took a look; another description of the underlying problem is that jQuery sortable and ember both want to manipulate the DOM. His solution is to cancel jQuery so Ember can do its thing: https://github.com/Addepar/ember-table/pull/265. I think we'll go with this approach, which should still fix your issue. Let us know if you have any objection!

mariuserik:
Awesome, LGTM.

PR #266, opened 2015-01-23
Make sure the element is sortable/resizable before destroying them.

liuyang-li:
@cyril-sf @yayalice


azirbel:
This makes sense to me, but I don't know how to judge if it is the best way. It seems especially strange to go to the controller to see if an element is sortable, rather than checking the element itself (e.g. is `.sortable` defined?).

It would be great if @ebryn or @cyril-sf could look at it.

cyril-sf:
calling `destroy` is the way to go. It's not clear to me if there is a better way to detect if the widget has been applied to an element than to check for a css class.

This seems brittle. But at least it works.

By looking at the source code of `jquery.ui.sortable.js`, it seems like there is a `ready` flag. Something to check maybe.

azirbel:
Sounds like Cyril is ok with it too, so LGTM.

liuyang-li:
Thanks for the comments. I pushed the change as it is for now to unblock me for another change. Agreed that we should be looking into a long term solution to this.

cyril-sf:
It definitely looks sad, but I haven't seen a better way to get the information.

PR #268, opened 2015-01-27
Reorder and resize fixes

azirbel:
Fixes https://github.com/Addepar/ember-table/issues/243.

Fixes a bug where reordering columns in fluid mode would cause errors.

Also fixes bugs around resizable handles for columns, especially in fluid mode.

Finally, fixes the configurable column demo to resolve a conflict with prettyprint DOM manipulation.

See commit messages for more details.

@Addepar/uie @liuyang-li @ebryn @Dylnuge


src/component.coffee
```diff
@@ -227,7 +226,7 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   # they can be. Note that this may fail to arrive at the table width if the
   # resizable columns are all restricted by min/max widths.
   doForceFillColumns: ->
-    allColumns = @get('allColumns')
+    allColumns = @get('columns')
```
Dylnuge: Looks like we don't use `allColumns` anywhere anymore; delete line 157 of this file?

Dylnuge:
LGTM

src/component.coffee
azirbel: deleted - thanks

PR #273, opened 2015-02-10
Refactor references to the component in views.

azirbel:
Previously we relied on `controller` to access the table component from within views.

This PR adds a mixin that looks up the table component via `nearestWithProperty`.

This is a cleaner way to do it (`controller` is not recommended as it doesn't support hierarchy well), and has the added benefit that you can pass in `tableComponent` if you want to separate the table component and the views.

@Addepar/uie


derekbrown:
LGTM.

dependencies/ember-addepar-mixins/resize_handler.js
```diff
@@ -36,6 +36,7 @@ Ember.AddeparMixins.ResizeHandlerMixin = Ember.Mixin.create({
     return this._setupDocumentHandlers();
   },
   willDestroyElement: function() {
+    Ember.run.cancel(this.endResize);
```
ebryn: I'd like to discuss `endResize` with you, it seems a bit odd

azirbel:
@ebryn after looking at that `resizeEnd` stuff I concluded that my change wasn't working, and that code is a terrible mess that I don't want to touch ([this in particular](https://github.com/Addepar/ember-table/blob/master/dependencies/ember-addepar-mixins/resize_handler.js#L14)) - so I just got rid of that commit.

PR #275, opened 2015-02-13
Make sure not to reorder fixed columns by mistake

azirbel:
Previously reordering columns would interfere with fixedColumns. This change compensates for the number of fixed columns.

Before recent changes, `onColumnSort` operated on `tableColumns` rather than `columns`. However, that caused bugs where the changes to `tableColumns` wouldn't update `columns`. The way it's done here is more one-directional flow (`tableColumns` is a derivative of `columns`; update `columns` directly).


Dylnuge:
LGTM

PR #276, opened 2015-02-17
Fix broken HTML

joeruello:
Fixes a few broken tags in the Docs found while building for HTMLBars. Thought I would contribute them back upstream.


azirbel:
Rebased and pushed to master as https://github.com/Addepar/ember-table/commit/a775ad2492a5c9a8d1a23a81a0532cb5a313c290.

Thanks @joeruello!

PR #277, opened 2015-02-17
Make `isSelected` check its row contents, rather row objects

philpee2:
Previously, `isSelected` checked if `_selection`, which contains row objects,
included a given row. Now, we use `selection`, which contains table content,
and check if it includes the row's content.

This means that when bodyContent is recomputed to contain new rows,
selection does not change, since selection is now determined based on
content, rather than rows.

@azirbel


azirbel:
LGTM

PR #279, opened 2015-02-18
Fix single selection mode.

azirbel:
Previously, single selection mode's `selection` output was an object, rather
than an array containing that object (which is be consistent with multiple
selection, and also what the docs say).

The previous change to `isSelected` meant that tables were broken entirely for
`selectionMode="single"`. This commit fixes that bug as well.

@philpee2 @Addepar/uie


philpee2:
LGTM

PR #280, opened 2015-02-20
Remove `_selection` and use content instead of rows in selection logic

philpee2:
@Addepar/uie especially @azirbel


src/component.coffee
```diff
@@ -381,55 +374,61 @@ Ember.AddeparMixins.ResizeHandlerMixin,
...
-      @get('persistedSelection').addObject row
+      @get('persistedSelection').addObject item
     else
       if event.shiftKey
         @get('rangeSelection').clear()
 
         lastIndex = @rowIndex(@get('lastSelected'))
+        # If the last selected row is no longer in the table, use the
+        # first row in the table
+        lastIndex = 0 if lastIndex is -1
```
azirbel: just questioned this, checked against finder, and finder does the same thing as the code above. :+1:

azirbel:
This is fantastic and I'm REALLY happy about it. LGTM.

src/component.coffee
philpee2: Nice!

PR #281, opened 2015-02-23
Switch back to old API for `selection`

philpee2:
A previous commit (88498d41f0356c8d2bf364882f25bc8723f6af5b) changed the
`selection` API so that it was always an array, even if `selectionMode` is
`single` or `none`. This commit changes back to the previous state.
  * When `selectionMode` is `none`, `selection` is `null`
  * When it is `single`, it is the selected item
  * When it is `multiple`, it is an array of selected items

Paired with @azirbel 

@Addepar/uie


azirbel:
I forgot - can you update the docs to explain all of this behavior?

src/component.coffee
```diff
@@ -77,12 +77,21 @@ Ember.AddeparMixins.ResizeHandlerMixin,
   # An array of the rows currently selected. If `selectionMode` is set to
   # 'single', the array will contain either one or zero elements.
   selection: Ember.computed (key, val) ->
+    mode = @get 'selectionMode'
```
azirbel: let's call this `selectionMode` instead, throughout this PR

philpee2:
Updated the docs and changed the variable name

azirbel:
LGTM

PR #283, opened 2015-02-26
Fix header height when table is resized

pbenes:
When resizing the browser window, table header row height was not updated
correctly. This shows when columns are autofilled to the complete table
width.

Sample jsfiddle with current behaviour:
http://jsfiddle.net/ghdyyfy9/

jsfiddle with the fix applied:
http://jsfiddle.net/kh07sff2/


azirbel:
This looks great - thanks @pbenes. Also, great PR - really like the before/after example.

LGTM

PR #284, opened 2015-02-26
Round widths on column resize

azirbel:
Previously, `onColumnResize` would be called with floating-point numbers, which
would then be used for a while until `doForceFillColumns` coerced them to be
integers.

Instead, we should just make them integers immediately.

@Addepar/ui2-developers


sreenath-addepar:
LGTM

Dylnuge:
LGTM2

PR #303, opened 2015-04-23
Upgrade to Ember 1.11.3

joeruello:
Me again, updated branch to support Ember 1.11.3. 

This is to avoid  `"Failed: The template being rendered by ... was compiled with `Ember@1.11.1` which does not match `Ember@1.11.3` (this revision)"` thrown when using the debug version of Ember.

Might be worth having different branches for each minor release as well as major releases?


azirbel:
Merged via https://github.com/Addepar/ember-table/commit/72130628efa04df71929a7f0cf3bafb995d29d61. Thanks @joeruello!

Maintaining these extra branches is also a lot of maintainance work (or will be, once we start releasing new versions). I don't think we have the time to maintain minor releases as well =/

joeruello:
Other option is once like CLI add-on is ready, refer people to using https://github.com/ef4/ember-giftwrap. I'm using it for liquid-fire and its working great.

PR #305, opened 2015-04-23
Proposed fix for #304

### LINKED ISSUE #93: Antiscroll and Bootstrap 3
I don't know if this is a bug or merely a matter for documentation, but I found that when the parent element of {{table-component...}} is of class "container", antiscroll doesn't render right. The vertical scrollbar is hidden beneath it's parent.

I didn't see this mentioned in any conversations here, so I thought I'd mention it.

### LINKED ISSUE #304: Scroll bar hidden due to column padding with bootstrap 3 (regression)
This is a regression of #93 

Latest release is back to "outerHeight" which re-introduces this issue.
One way to fix this would be to wrap the entire table in an extra div with no paddings/margins, but it seems more appropriate to respect padding and margins on any existing container by using .height (not .innerHeight as in original fix) instead.

drglitch:
- Fixes margin and padding of parent being included in container/row calculation
- closes #304, re-closes #93
- allows table to play nice with bootstrap and similar frameworks (where parent Cols may have margin and/or padding)

### DIFF
--- a/src/component.coffee
+++ b/src/component.coffee
@@ -194,8 +194,8 @@ Ember.AddeparMixins.ResizeHandlerMixin,
 
   elementSizeDidChange: ->
     return unless (@get('_state') or @get('state')) is 'inDOM'
-    @set '_width', @$().parent().outerWidth()
-    @set '_height', @$().parent().outerHeight()
+    @set '_width', @$().parent().width() # fixing margin problems see issue #304
+    @set '_height', @$().parent().height()
     # we need to wait for the table to be fully rendered before antiscroll can
     # be used
     Ember.run.next this, @updateLayout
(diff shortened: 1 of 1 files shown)


azirbel:
Sorry for the regression @drglitch, and thanks for the fix. I'll check it out later today.

I think there may have been a reason we changed it back to `outerWidth`; I want to see if I can find out why.

drglitch:
Thought the same - the original fix proposed in #93 was not entirely accurate as it suggested .innerWidth which includes padding (https://api.jquery.com/innerWidth/) - so maybe someone saw that as a problem and regressed it.

I believe that .width is best choice here as that respects entirety of CSS styling on parent.

PS: travisCI is failing due to missing config :)

azirbel:
Ok, this looks great - will merge now. Thanks for your excellent explanations @drglitch! Also, thanks @lanceaustin26; it looks like I missed your comment in #93 the last time around, but I found it really helpful while looking back at this issue.

azirbel:
Merged as https://github.com/Addepar/ember-table/commit/bb09757c188f5185a6b789a983b764cdd2cf1a9f.

drglitch:
@azirbel awesome - thanks! :)

PR #306, opened 2015-04-23
Move back to using pointer style sortable tolerance from intersect

Dylnuge:
Intersect style tolerance for UI sortable (reorderable) elements in JQuery
produces a bug where we can't drag columns to the leftmost slot unless we grab
them from the far left side.

@azirbel and I worked together on this

@ldevalliere any memory of why e1bc1d6fb was made? This behavior feels like the opposite of what I'd expect `pointer` and `intersect` to do, and I'm confused.


ldevalliere:
"Pointer" is what we wanted, however, "intersect" worked slightly better (but still buggy). We couldn't reorder a column into the last column position with "pointer". This is also connected to the unfortunate 3px we add on to "_tableColumnsWidth" to make column ordering into the last column possible. I am working on that issue on https://github.com/ppongzenefits/ember-table-addon/tree/dev

Dylnuge:
Neither quite behaves the way I'd expect them too. Here's the oddities I can observe with both:

### 'pointer' tolerance
* Can't drag a column larger than all the columns before it to the left if you grab on the right side of the column
* Dragging a column all the way to the right puts it in the penultimate position. Dragging it a little less than all the way still puts it last.

### 'intersect' tolerance
* Strangely, where you grab the header still matters. Grabbing headers on the rightmost side makes it impossible to drag them to the leftmost position.

azirbel:
Here's the SO post we found earlier: http://stackoverflow.com/questions/10637095/jquery-ui-sortable-tolerance-option-not-working-as-expected. This is pretty frustrating, but I think it might actually be worth doing something custom if possible. It's pretty low-priority though. For now I think we should just change it back to "pointer"; it subjectively feels a bit less buggy to me.

Dylnuge:
I agree with the subjective 'pointer' feeling more right than 'intersect' (except for that dragging all the way right thing in fixed width tables...that's just weird).

Dylnuge:
Since @azirbel and I did this together (as super minor as it is), can someone else review? @yayalice or @philpee2 maybe?

philpee2:
LGTM

Issue #165, opened 2014-06-08
isLastRow property breaks ajax lazy loading tables


jrhe:
When doing something along the lines of the Ajax example, because of the isLastRow property, the very last element of content is requested. This subsequently requests the last page of data.

Perhaps a better way to do the isLastRow functionality would be to store an index of the row and compare it to the length? Willing to do a pull request for this if its a good way to go.

Leooo:
Hi, is this issue still current? I'm a third-day user of Ember and want to try ember-table, but the ajax example doesnt render anything (ajax simple was fine), I think this is because of this thread's issue (the index of ObjectAt in the controller starts with the last row, not the first one). Also, the name of this thread maybe should be updated to "ajax cells.." for those who are trying to understand why the examples don't work? Just some suggestions of a newbie, but everyone is a newbie at the beginning..

azirbel:
Hi @Leooo - yep, this is still current. However, the ajax example (http://addepar.github.io/ember-table/#/ember-table/ajax) does work - the issue described here is that ember-table requests the last row of data earlier than necessary. If you're having trouble with the AJAX table, please open a new issue or stack overflow question!

Leooo:
OK thanks for precising this, I forgot to precise I was using the Ember Table ember-1.10 branch with ember 1.10 / htmlbars. If this should work then I must have done something wrong on my side, will work more on this.

Leooo:
Actually now that I'm able to see exceptions in the console I have this one when starting ember-table ajax example:
"TypeError: createChildViews first argument must exist"

I checked that the patch of @gunn referenced in https://github.com/ember-cli/ember-cli/issues/1331 is in my ember-table.js file, so that's not the problem.

Another thing I tried is to rename my app name  from window.App to window.AssessUp, as this seemed to be possibly an issue following the related threads.

Will investigate more, just posting this in case you are familiar with this exception on ember-table. I will be so happy to finally be able to get rid of jquery-datatable which really doesnt marry well with large data sets streamed from the server..

I will create another issue or SO in the next days if I can't solve this.

Using ember-1.10 branch with htmlbars

Leooo:
Got it working by removing the avatar column. Not opening another thread as this is solved on my side. Thanks.

Leooo:
[...]

PR #311, opened 2015-05-21
avoid access last row.

cyjia:
Set lastItem on RowArrayController, to address issue #165


azirbel:
I see you commented on https://github.com/Addepar/ember-table/issues/165 - adding a note that this fix is meant to address that issue.

app/controllers/simple.coffee
```diff
@@ -36,3 +36,14 @@ App.EmberTableSimpleController = Ember.Controller.extend
       close: Math.random() * 100 - 50
       volume: Math.random() * 1000000
   .property 'numRows'
+
+  actions:
```
azirbel: we shouldn't add this to a simple example, and probably shouldn't change the demo app in general.

azirbel: I guess you are making it easy for us to test your change; just need to remember to remove these before we merge.

src/utils/utils.coffee
```diff
@@ -72,14 +72,19 @@ Ember.Table.RowArrayController = Ember.ArrayController.extend
...
     subController = subControllers[idx]
 
+    @set('lastItem', subController) if @isLastItem(idx)
     return subController if subController
     subController = @get('itemController').create
       target: this
       parentController: @get('parentController') or this
       content: object
     subControllers[idx] = subController;
+    @set('lastItem',subController) if @isLastItem(idx)
```
azirbel: it would be great if we could restructure this logic such that `@set('lastItem',subController) if @isLastItem(idx)` only needs to be called once. Something like:
```
unless subController
  subController = @get('itemController').create
    target: this
    parentController: @get('parentController') or this
    content: object
  subControllers[idx] = subController;
@set('lastItem',subController) if @isLastItem(idx)
```
There should also be a comment explaining why we need to do this.

src/utils/utils.coffee
```diff
@@ -72,14 +72,19 @@ Ember.Table.RowArrayController = Ember.ArrayController.extend
...
     return subController;
 
+  isLastItem: (idx) ->
```
azirbel: I'd prefer calling this `_isLastItem` to make it clear it's for private use.

src/utils/utils.coffee
```diff
@@ -72,14 +72,19 @@ Ember.Table.RowArrayController = Ember.ArrayController.extend
...
+    return @get('content').length - 1 == idx
```
azirbel: `===`

azirbel:
This is a clever approach; I like it.

@jrhe if you are still interested in a fix, can you confirm that this works for you?

I think it could use a comment or two in the code to make it clear what's going on, since this is a bit of a hack. (It is already very strange that we override `controllerAt`).

azirbel:
It will also need to be rebased on the new Ember CLI version of Ember Table.

azirbel:
Moved to https://github.com/Addepar/ember-table/pull/320.

PR #317, opened 2015-05-25
Remove demo app's dependency on lodash

azirbel:
This makes the examples easier to use - you can copy+paste the simple
table example into a new project and it will work immediately.

@Addepar/ui2-developers

### DIFF
--- a/tests/dummy/app/controllers/bars.js
+++ b/tests/dummy/app/controllers/bars.js
@@ -22,15 +22,17 @@ export default Ember.Controller.extend({
   }),
 
   tableContent: Ember.computed(function() {
-    return _.range(100).map(function(index) {
-      return {
-        key: index,
+    var content = [];
+    for (var i = 0; i < 100; i++) {
+      content.pushObject({
+        key: i,
         value1: Math.random() * 80 + 10,
         value2: Math.random() * 80 + 10,
         value3: Math.random() * 80 + 10,
         value4: Math.random() * 80 + 10,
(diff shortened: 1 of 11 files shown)


tests/dummy/app/controllers/bars.js
```diff
@@ -22,7 +22,7 @@ export default Ember.Controller.extend({
   }),
 
   tableContent: Ember.computed(function() {
-    return _.range(100).map(function(index) {
+    return Array.apply(null, new Array(100)).map(function(_, index) {
```
cyril-sf: Why using `Array.apply`?

azirbel: maybe a `for` loop would have been fine... is that what you mean? I was going for a drop-in substitution but now I'm think a for loop would have been better.

cyril-sf: I didn't even realize that we were only using it to iterate.

At the very least,

```js
var content = new Array(100);
return content.map(
    ...
);
```

should be enough.

azirbel: Can't do that unfortunately, see http://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-an-array-based-on-suppl. But we could use a for loop to make it easiest to read. Since they're examples, I think I will do that.

azirbel:
@cyril-sf I have updated it to use for loops.

cyril-sf:
LGTM

PR #318, opened 2015-05-25
Add "grunt clean" task to clear temp files

azirbel:
@Addepar/ui2-developers

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -5,6 +5,7 @@ module.exports = function (grunt) {
 
   grunt.loadNpmTasks('grunt-banner');
   grunt.loadNpmTasks('grunt-broccoli');
+  grunt.loadNpmTasks('grunt-contrib-clean');
   grunt.loadNpmTasks('grunt-contrib-uglify');
   grunt.loadNpmTasks('grunt-release-it');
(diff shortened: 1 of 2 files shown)


cyril-sf:
Doesn't Ember-CLI expose something to do it?

azirbel:
Looks like there is https://www.npmjs.com/package/ember-cli-clean but nothing else.

azirbel:
And that one reinstalls, which is unexpected. (https://github.com/technomage/ember-cli-clean/blob/master/lib/command.js#L13) I prefer using this grunt task.

Gruntfile.js
```diff
@@ -25,6 +26,8 @@ module.exports = function (grunt) {
       }
     },
 
+    clean: ['tmp', 'ember-dist', 'node_modules', 'bower_components'],
```
cyril-sf: do we really want to remove all packages and not only our build files?

azirbel: I guess this PR is mostly a convenience for me. I find myself `rm`'ing these folders a lot, so I can reinstall and rerun from scratch. I mostly use it when we update dependencies.

cyril-sf:
LGTM

PR #319, opened 2015-05-25
Port commits from ember-table-addon

azirbel:
The last three commits in `ember-table-addon` weren't pulled into #309. I'm cherry-picking them back in.

@Addepar/ui2-developers


azirbel:
Technically this doesn't need review since I've looked over it myself. But it doesn't hurt; maybe I missed something.

azirbel:
Also apparently `styleString.htmlSafe().toString();` isn't useful. I haven't checked it myself.

azirbel:
Merged.

PR #320, opened 2015-05-27
fix issue #165, isLastRow property problem.

cyjia:
- isLastRow property will cause ember table to access last row of content, see issue #165 
  - create an new property 'lastItem' to replace calling of `tableComponent.bodyContent.lastObject`


azirbel:
This seems to work well - thanks @cyjia! Cleaned up comments and merged via https://github.com/Addepar/ember-table/commit/d464dada6e50e0949be1b8abda756108d94d1020.

PR #325, opened 2015-06-02
Update .npmignore to prevent packaging tmp dir

jimnanney:
The ember-cli addon blueprint generates the .npmignore and previous to 0.2.2 did not ignore the tmp directory causing large packages. This change is direct from https://github.com/ember-cli/ember-cli/pull/3539

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 bower_components/
 tests/
+tmp/
 
 .bowerrc
 .editorconfig


jimnanney:
Should dist/ and .stage/ be added in the npmignore as well?

azirbel:
LGTM, thanks! Merged via https://github.com/Addepar/ember-table/commit/ec07f50e52a7a5c394ba6e95c56e268d74faca1c. Opened https://github.com/Addepar/ember-table/pull/339 for the other two.

PR #329, opened 2015-06-09
adds element resize detection

jrhe:


billy-addepar:

PR #332, opened 2015-06-11
Fix memory leaks with antiscroll

cyril-sf:
@Addepar/uie 

@yayalice @joseph-chu


Dylnuge:
LGTM

Dylnuge:
Wait where does src/component.coffee come from? I think you can't merge this into the new ember-cli app.

cyril-sf:
You're right, I actually don't want to merge it. This fix is already in the Ember CLI addon, but we don't use it yet.

PR #335, opened 2015-06-12
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3

Also their maintainer says they probably won't ever use it: http://stackoverflow.com/questions/24844901/bowers-bower-json-file-version-property

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,5 @@
 {
   "name": "ember-table",
-  "version": "0.5.0",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


azirbel:
Merged via https://github.com/Addepar/ember-table/commit/4d6de7e998d5a38860c43aca4e36698ffa76fa3a; modified the commit to work with the release process as well.

PR #340, opened 2015-06-19
Fix lazy container bug

jrhe:
When changing content of ember-table I end up with an error stating
that I can't call get on undefined. This stems from a problem with
lazy container where the container views length is longer than the
actual number of views. I'm unsure if this is an ember problem or if
its something to do with the implementation of lazy container itself.

It might be that you can't rely on ember having created all the views.
Checking to see if the view exists before settings its content fixes
this and I have had no noticeable problems with this extra check.

screen shot 2015-06-19 at 19 49 41

### DIFF
--- a/addon/views/lazy-container.js
+++ b/addon/views/lazy-container.js
@@ -82,7 +82,7 @@ StyleBindingsMixin, {
       var itemIndex = startIndex + i;
       childView = childViews.objectAt(itemIndex % numShownViews);
       var item = content.objectAt(itemIndex);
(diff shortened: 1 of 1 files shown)


azirbel:
Merged via https://github.com/Addepar/ember-table/commit/fa6712392aa935110c731d5ca16c718992a0513f. This is pretty weird, but in this case I don't see any harm in checking for `childView`, since the content of the `if` block only affects `childView` anyway.

PR #341, opened 2015-06-22
Selection action

jrhe:
Add an action which gets fired when the selection changes with the current selection. 
May want to deprecate two way selection binding to adhere more to the Data down, actions up pattern.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -105,6 +105,10 @@ StyleBindingsMixin, ResizeHandlerMixin, {
     }
   }).property('persistedSelection.[]', 'rangeSelection.[]', 'selectionMode'),
 
+  _selectionChanged: Ember.observer('selection', function() {
+    this.sendAction('selectionChanged', this.get('selection'));
+  }),
+
   // ---------------------------------------------------------------------------
   // Internal properties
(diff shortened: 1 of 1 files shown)


addon/components/ember-table.js
```diff
@@ -105,6 +105,10 @@ StyleBindingsMixin, ResizeHandlerMixin, {
     }
   }).property('persistedSelection.[]', 'rangeSelection.[]', 'selectionMode'),
 
+  _selectionChanged: Ember.observer('selection', function() {
```
azirbel: should observe `selection.[]`, right?

jrhe: Correct. No idea why/ how this is working in  my app.

azirbel:
I'm in favor of adding this action, but (1) I think it should be called `onSelectionChanged` to be consistent with https://github.com/Addepar/ember-table/pull/350, and (2) I'm not convinced using an observer is the right way to do it. This means e.g. if you use the two-way binding to set `selection` from outside the table, the observer will fire. But the action should only be triggered when table-internal logic makes the selection change. Maybe this `sendAction` call belongs in `setSelected` (and may have to be added elsewhere as well).

billy-addepar:

Issue #31, opened 2013-02-28
Filtering, sorting, grouping, and column showing / hiding


alexspeller:
I've been implementing filtering, sorting, grouping, and column selection for a project. How interested are you in any / all of these features being a part of the core of ember-table?

I'd be happy to refactor my code into a pull request if you are interested in it, but I don't want to waste effort if you don't think these features should be in the core library.

 In terms of code, it would be about ~200 lines of coffescript, ~50 lines of handlebars and ~50 lines of css altogether for all of the features, so it's quite lightweight.

Please let me know your thoughts.

billy-addepar:
This version of Ember table is no longer supported. If you want to continue discussion, you can open the issue on [Ember Table Legacy](https://github.com/Addepar/ember-table-legacy)

PR #342, opened 2015-06-22
Adds onColumnSort action

jrhe:
Fire an action when columns are sorted.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -177,6 +177,7 @@ StyleBindingsMixin, ResizeHandlerMixin, {
     columns.removeObject(column);
     columns.insertAt(numFixedColumns + newIndex, column);
     this.prepareTableColumns();
+    this.sendAction('onColumnSort', columns);
   },
 
   // An array of Ember.Table.Row computed based on `content`


azirbel:
The function `onColumnSort` is intended to be a hook you can override to get this behavior. I agree that we should add an action that fires here, but think we should give it a different name. It's also a good opportunity to stop calling this a "sort" and start calling it a "reorder".

How about `this.sendAction('columnsReordered', columns);`?

jrhe:
The problem with columnsReordered is that its not obvious that its expecting an action. Couldn't actually get onColumnSort to work. Whatever we choose it should be consistent with other actions in the component so we should consider what other actions might be useful.

azirbel:
I don't have strong opinions on the naming conventions for actions. What name do you prefer? I agree we should be consistent about actions as we start to introduce them.

Other actions we'll need:
* `selectionChanged`: https://github.com/Addepar/ember-table/pull/341

When using the component, it will look like this:

```
{{ember-table
  columns=tableColumns
  content=tableContent
  columnsReordered="handleColumnReorder"
}}
```

Based on that, maybe `onColumnReorder` is a better action name?

@Dylnuge @cyril-sf @igillis

drglitch:
@azirbel I highly vote for "onColumnsReordered" or similar (notice plural too). onColumnSorted is something that should happen when a particular single column gets sorted (i.e. data in it, like being discussed in https://github.com/Addepar/ember-table/issues/31) ... so if in future ember table core supports sorting, this would become ambiguous.

azirbel:
Merged https://github.com/Addepar/ember-table/commit/60baf95769366f8f3a78811d38236405a0be9fef, which accomplishes the same thing.

PR #344, opened 2015-06-26
Fixed bug scrollbar always appears

thangdinh:
Problem: Scrollbar always appears because of hacky padding is added 
but not subtracted when calculating available width
Solution: Subtracted that hacky padding number when calculating the available width
@azirbel @Dylnuge @cyril-sf


azirbel:
Previous work on this issue: https://github.com/Addepar/ember-table/pull/126 and https://github.com/Addepar/ember-table/pull/172. Based on past experience, I think this solution is probably missing one of (1) ability to reorder columns to the rightmost column or (2) scroll still works after table resize. Will look at it more later.

thangdinh:
@azirbel I guess it won't be an issue for the 2 abilities that you mentioned. I checked #172 and #126 but I don't find it very relevant. One is trying to remove the hacky padding, while mine is trying to update the right content-size so that the scrollbar won't appear by default because of that 3px padding.
But I'm not 100% how it works. I just spotted on and went through the code a little bit to understand the problem and tried to fix it, so please let me know if it doesn't help.
But in any case, if we can get rid of that hacky padding, it will be great.

thangdinh:
I'll close it for now as it doesn't resolve the issue completely.

PR #345, opened 2015-07-01
Header aligns properly after reordering columns

dianeyang:
Before, you could cause the column headers to get out of alignment with the rest of the table by scrolling a bit then reordering the columns (for example, here: http://addepar.github.io/ember-table/#/financial)

This change makes it so that sidescrolling gets triggered when the columns change. This also removes references to scrolling from header rows (as they weren't actually doing anything, and scrolling should be handled in the header block).

@azirbel @Dylnuge @cyril-sf @philpee2


addon/views/header-block.js
```diff
@@ -9,5 +9,12 @@ export default TableBlock.extend({
 
   content: Ember.computed(function() {
     return [this.get('columns')];
-  }).property('columns')
+  }).property('columns'),
+
+  onColumnsDidChange: Ember.observer(function() {
+    var _this = this;
+    Ember.run.schedule('afterRender', function() {
```
Dylnuge: May need to save this to cancel in the `willDestroyElement` function.

Dylnuge:
LGTM besides comment

addon/views/header-block.js
Dylnuge: Not 100% sure but pretty sure this will work without, since it's always in the same run loop and should happen before a `destroy`

PR #346, opened 2015-07-06
Export and search functionality

ManojrajT:
I have used this ember-table now i need to export the table to excel, pdf etc format along with search functionality.Is such functionality exist if is there hen redirect me to specific link. If anyone has developed it then please share.


azirbel:
Hi @ManojrajT. Ember Table doesn't have these features (and export to pdf/excel is definitely out of the scope of the project). You'll need to write these yourself, and your solution won't involve Ember Table - it'll be an operation on the underlying `content` array that is backing the table.

In the future, questions like this are better asked on Stack Overflow; our GitHub page is used for reporting bugs and discussing features for Ember Table. This also should have been an issue (https://github.com/Addepar/ember-table/issues) rather than a pull request, since pull requests are used for submitting new code to the repository. Finally, in the future please do more research before asking this kind of question; it's clear from our demo page (http://addepar.github.io/ember-table/#/overview) that these features are out of scope.

PR #347, opened 2015-07-06
Upgrade to EmberCLI 0.2.7

taras:
- Upgraded dependencies to support EmberCLI 0.2.7
- Replaced 6to5 with babel.js
- Moved JavaScript template for Configurable Table to a component because template parser was choking on JS syntax


azirbel:
Moved to https://github.com/Addepar/ember-table/pull/354, let me know what you think!

PR #349, opened 2015-07-08
Add acceptance tests

taras:
This branch is rebased on EmberCLI 0.2.7 branch.

## Includes
- Acceptance Tests
   - Simple Test
   - Configurable Columns Test
   - Removable Columns Test
- Refactored template substitution with `ember-code-snippet`
- Refactored all examples to use new `{{code-snippet}}`
- Added random module with `randomNumber` and `randomDate` that produce predicable values based on seed value
- Added `rowText` test helper for getting cell text for a row


taras:
@azirbel I rebased this branch against #354

tests/dummy/app/components/configurable-table-source.js
```diff
@@ -0,0 +1,34 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
```
azirbel: do you really want to make this a component? it's not needed after the changes on master

taras: No, I don't like it. let me see if I can revert it.

taras:
I reverted that change

tests/dummy/app/templates/ajax.hbs
```diff
@@ -5,7 +5,17 @@
     <div class="col-md-12">
       <div class="example-container">
         <div class="ember-table-example-container">
-          {{partial "ajax-table/ajax-table"}}
+          {{!// BEGIN-SNIPPET ajax-table}}
```
azirbel: YESSS

tests/helpers/ember-table.js
```diff
@@ -0,0 +1,8 @@
+import Ember from 'ember';
+
+Ember.Test.registerHelper('rowText', function rowText(app, rowIndex=0){
+  let cells = app.$(`.ember-table-table-row:eq(${rowIndex}) .ember-table-cell`);
```
azirbel: why backticks?

tests/dummy/app/utils/random.js
```diff
@@ -0,0 +1,25 @@
+var seed = 6;
+// source: http://indiegamr.com/generate-repeatable-random-numbers-in-js/
+// in order to work 'Math.seed' must NOT be undefined,
+// so in any case, you HAVE to provide a Math.seed
+export function randomNumber(max = 1, min = 0, randomSeed = undefined) {
```
azirbel: why add `randomSeed` as the parameter here? I think it should be done separately, as it is currently done, via `setRandomSeed`.

azirbel: never mind - now I see the usage a few lines down.

tests/acceptance/configurable-columns-test.js
```diff
@@ -0,0 +1,29 @@
...
+    Ember.run(application, 'destroy');
+  }
+});
+
+test('table renders', function(assert) {
+  visit('/configurable-columns');
+  andThen(function() {
+    assert.equal(currentPath(), 'configurable-columns');
+    assert.deepEqual(rowText(0), ['Date', 'Open', 'High', 'Low', 'Close'], "headers");
+    assert.deepEqual(rowText(1), ['Thu Jul 14 2005', '22.49', '-18.01', '-47.10', '-0.30'], "values in first row");
```
azirbel: break into two lines

tests/acceptance/removable-columns-test.js
```diff
@@ -0,0 +1,60 @@
...
+    Ember.run(application, 'destroy');
+  }
+});
+
+test('table renders', function(assert) {
+  visit('/removable-columns');
+  andThen(function() {
+    assert.equal(currentPath(), 'removable-columns');
+    assert.deepEqual(rowText(0), ['Date', 'Open', 'High', 'Low', 'Close'], "headers");
+    assert.deepEqual(rowText(1), ['Thu Jul 14 2005', '22.49', '-18.01', '-47.10', '-0.30'], "values in first row");
```
azirbel: break some of these into 2 lines

azirbel:
This is so awesome, @taras. LGTM after a couple recent comments about line length.

It's too bad that we have to use `randomNumber`, because it makes the examples not work when copy+pasted. But I think it's worth it to make things testable. We can change that later if we change our minds. Another alternative would be to use more predictable content (ascending numbers?).

Could you squash into a single commit and make sure `dist` files are updated? Then I'll merge.

tests/helpers/ember-table.js
taras: Backticks are ES6 template strings that are transpiled by babel.js to ```'.ember-table-table-row:eq(' + rowIndex + ') .ember-table-cell'

taras:
@azirbel I think we can stub Math.random in tests and go back to use it. I'm not sure about Date because there is no Date.random. With that said, I'm not sure how much we should focus on making the examples copypastable especially the `tableContent`. 

I think for the benefit of people who're going to start using Ember Table, we should try to make the examples look as close to real life as possible, include refactoring the examples to use routes to provide data rather than controllers.

I applied the line changes that you suggested, regenerated the dist files(which didn't change cause I regenerated them previously) and squashed into 1 commit.

Should be good to go.

Issue #312, opened 2015-05-21
Event for column re-ordering


catc:
Is there an event that fires when columns are re-ordered? I think it'd be useful in helping add additional customizability to table. 

I'm trying to set it up so when a user re-orders a column, an event fires triggering an action to save the new column order in local storage. Then on refreshing the page, table initiation checks local storage for column order before rendering the table.

azirbel:
There isn't an event that fires. Right now you have to add one yourself by extending Ember Table using `Ember.Table.EmberTableComponent.extend({...})`.

I've heard a few requests for this feature now, and I like your explanation of the use case. I think it would be worth adding an event like this to Ember Table so everyone can use it. PRs are welcome!

drglitch:
Just came here to request similar - event triggered when the columns are *resized*.
Right now, jquery ui resizable just calls the `onColumnResize` function (which i am catching and extending), but thats hacky. Having EmberTable emit a custom "i've just resized" event would be much cleaner.

The usecase is the following: A custom template inside HeaderTableCell that has some logic hooked onto it via `didInsertElement`. When the cell is resized, i need to either show or hide "overflow" indicator - which is a div inside the table-cell-container on that particular cell. having a custom event being would turn this into a simple `$('.myelem').on('ember-table-header-col-resized', function() {$(this).toggle();});`

azirbel:
I think that's a reasonable use case and would be happy to add it @drglitch. Also, see the discussion at https://github.com/Addepar/ember-table/pull/342 - do you have opinions on the naming convention for our actions?

PR is welcome.

azirbel:
Actually, this issue will be resolved by that PR, so I'll close this one. Open a separate issue/PR for the resizing issue and we can talk more about that.

drglitch:
@azirbal - dont think above PR in its current state closes this. it closes column reordering, but not resizing. The latter can easily be added in similar vain though.

azirbel:
Yep, I think we should add the latter in a new issue/PR :)

drglitch:
Reading comprehension is not my strongest trait apparently :) I'll submit a PR for this tomorrow.

PR #350, opened 2015-07-08
add onColumnReordered and onColumnResized actions

drglitch:
- related to #312 and supersedes #342
- onColumnReordered fires with three args: all columns (new), column
that was reordered, and index at which new column was placed
- onColumnResized fires with 2 args: column that was resized and new
width of the column. Note that resizing of a column can cause another
to resize if fit columns mode is on.

- manual tests passed. no automated tests ran.


taras:
@drglitch if you don't mind, could you please describe the use case where you need these actions?

drglitch:
@taras please see https://github.com/Addepar/ember-table/issues/312#issuecomment-119353690. i can go into more detail if necessary :)

taras:
@drglitch Got it. Thanks :)

azirbel:
Merged as https://github.com/Addepar/ember-table/commit/60baf95769366f8f3a78811d38236405a0be9fef. Thanks @drglitch!

PR #354, opened 2015-07-15
Upgraded to EmberCLI 0.2.7

azirbel:
Same as https://github.com/Addepar/ember-table/pull/347, but:

* Rebased on master
* Use Ember version `>=1.12 <1.13` instead of `1.12.0`
* No need to change the template for configurable table; the parser was choking on a `>` sign which I fixed in a previous commit on master.

@Addepar/ui2-developers @taras


azirbel:
Still need to check against JS bin and make sure that the globals build works.

taras:
LGTM :+1:

azirbel:
Globals build works, merge time

taras:
WOOP WOOP

PR #358, opened 2015-07-20
Improved antiscroll rebuild

taras:
- Added resize-container-test route that shows table that's resized when container size changes
- Added test that shows that scrollbar gets broken when resizing a table
- Added skipped test that shows an unnecessary scrollbar when rendering the table will `columnsFillTable=true`

PR #360, opened 2015-07-22
Added maxHeight to Ember Table

taras:
maxHeight makes it possible to specify maximum table on the table. This makes the height variable. The table will use available height until maxHeight is reached and then show vertical scrollbar.

Includes the following changes:
- PR #358 
- Configured travis to allow Ember 1.13+ to fail
- Upgrade to ember-qunit to allow integration tests
- Cleaned up `ember-table` component to remove `.property()` prototype extension use
- Added integration test for maxHeight functionality


billy-addepar:

PR #361, opened 2015-07-24
Ember 1.12 new CP syntax

artych:
To avoid ember 1.12 deprecation warning: 

     DEPRECATION: Using the same function as getter and setter is deprecated. 
     See http://emberjs.com/deprecations/v1.x/#toc_deprecate-using-the-same-function-as-getter-and-setter-in-computed-properties for more details.


addon/components/ember-table.js
```diff
@@ -81,29 +81,34 @@ StyleBindingsMixin, ResizeHandlerMixin, {
...
+        this.get('persistedSelection').clear();
+        this.get('rangeSelection').clear();
+        switch (selectionMode) {
+          case 'single':
+            this.get('persistedSelection').addObject(val);
+            break;
+          case 'multiple':
+            this.get('persistedSelection').addObjects(val);
+        }
+      }
```
azirbel: for consistency, this should `return this.get('selection')` or duplicate the code.

addon/components/ember-table.js
```diff
@@ -117,12 +122,13 @@ StyleBindingsMixin, ResizeHandlerMixin, {
...
+  _resolvedContent: Ember.computed('content', {
+    set: function(key, value) {
       return value;
-    } else {
-      var _this = this;
-      value = [];
+    },
+    get: function() {
+      var _this = this,
+          value = [];
```
azirbel: should declare with separate `var` statements

addon/components/ember-table.js
```diff
@@ -191,13 +197,15 @@ StyleBindingsMixin, ResizeHandlerMixin, {
   }).property('_resolvedContent.[]'),
 
   // An array of Ember.Table.Row
-  footerContent: Ember.computed(function(key, value) {
-    if (value) {
-      return value;
-    } else {
-      return Ember.A();
+  footerContent: Ember.computed({
+    set: function(key, value) {
```
azirbel: to mimic the old code, this looks like it should be:
```
set: function(key, value) {
  return value;
},
get: function() {
  return Ember.A();
}
```
But I'm not actually sure what the old code is doing here.

azirbel:
Multiple selection mode also doesn't seem to work.

Otherwise, I like the work in this PR. I'm excited to get rid of deprecation warnings :+1:

artych:
Thanks for feedback, I've fixed code and checked multiple selection mode on simple table. Looks like it works now.

PR #363, opened 2015-07-24
Add state inDOM check for an afterRender call

azirbel:
Fixes a bug introduced by https://github.com/Addepar/ember-table/commit/fa75c3932732554ab98d577d47a5f6680d37dfe0, where `_this.$()` can be undefined during teardown.

@dianeyang


dianeyang:
LGTM!

Issue #337, opened 2015-06-15
Ember 1.13 support


jrhe:
Currently ember-table breaks with 1.13. I expect this is mainly because of glimmer.

Notes from tinkering:

The first thing which breaks are references to tableComponent as it seems it hasn't been injected when it reaches certain references. Replacing this injection code with an alias to parentView seems to get things ticking a bit further. Why is this mechanism used for providing a reference to the table component? It seems very hacky.

After this references to `this.$()` fail. `$()` only seems to be available after render, which makes sense as it get the jquery element for the current component/view. These seem to be fixable by delaying till after rendering of the component using `Ember.run.schedule('afterRender', ...)`.

Doing these two things removes the errors but results in a non functioning table. The header row is rendered along with its contents but the contents aren't visible and I can't figure out why. The rows themselves aren't rendered.

jimnanney:
Got a fork of your work so far? I'm going to go down the 1.13 path soon, and anything I can do to make this easier would be a good thing.

shaunc:
@jrhe -- could the problem be the antiscroll stuff trying to read the "actual" dimensions of the thing its writing into, and not finding the element as it hasn't been drawn yet?

jrhe:
I don't have a fork as it was more exploratory hacks rather than anything all that useful. It seems that the widths dont get propagated down to the right views resulting in some being 0

goodwink:
I'm very interested in this as well as we're about to kick off an ember 1.13/2.0 effort.  If there's a fork or something we can contribute to in order to help I will have several people available in the next few weeks.

HenryVonfire:
Views won't be available in ember 2.0 and ember-table is completely base on views so there will be a lot of work ahead to switch from views to components. I've been playing around to see if I can do that easily but I'm facing a problem with the mixin register-table-component. Here is a schema of the things I've been able to convert into component successfully:

<img src="https://lh3.googleusercontent.com/fphJVeNPOmMCLDKbVFmd5aUGV6wipIdTzpY0Oczf-Uk_gymZG6TmUM5EKzbiF2s8PbPhW5uIlRSjT7o=w1888-h802">

joeruello:
@HenryVonfire Have you got a fork we can have a look at, happy to help.

@azirbel is there any preferred way for us to tackle this?

jrhe:
[...]

PR #364, opened 2015-07-27
Don't test with release/beta/canary versions of Ember

azirbel:
Ember Table will need major refactors to work with Ember 1.13, so
testing against these versions of Ember using `ember-try` doesn't make
sense.

@Addepar/ui2-developers @embersherpa

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,14 +11,9 @@ cache:
 
 env:
   - EMBER_TRY_SCENARIO=default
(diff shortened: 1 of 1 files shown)


jordansmith42:
LGTM makes sense, but this does seem like something we would want to run once we get ember table running on the current release.

azirbel:
@jordansmith42 unfortunately it may be a while: https://github.com/Addepar/ember-table/issues/337.

PR #365, opened 2015-07-27
Rendering changes related to force fill columns

taras:
- when forceFillColumns used, the columns should recalculate when table width changed - [diff](https://github.com/Addepar/ember-table/commit/1e2c5214dc14bd7001995f2112c9ff2a4aecd846)
- when scheduled force fill columns is called, make sure it's conditional - [diff](https://github.com/Addepar/ember-table/commit/d92554685a49b2dc0c1f9c0d107e2636e9590ed9)
- force fill columns should be executed before antiscroll - [diff](https://github.com/Addepar/ember-table/commit/52d8785c9351852803a310bbb562c7de5e128618)
- table width should be updated if table height caused page to show a scrollbar - [diff](https://github.com/Addepar/ember-table/commit/aa7747fe582ad835d378fb5e55d3f6446b106acd)


billy-addepar:

Issue #356, opened 2015-07-16
Consider/experiment with use supergroup.js for data mangling


taras:
Preparing data for presentation in the browser can be a real mess. On most projects that I've worked with, we usually start off with doing that in the browser and later moving it into the API. This simplifies the frontend code and makes everything much simpler to think about. But often, data mangling in the browser is unavoidable, especially for tables like the Financial Table Example.

It would be great if we had a standard way of preparing data to be presented in the table. For this purpose, we might want to look at using something like [supergroup.js](https://github.com/Sigfried/supergroup) which is designed for this purpose. 

Toptal has an article about [Ultimate In-memory Data Collection Manipulation with Supergroup.js](http://www.toptal.com/javascript/ultimate-in-memory-data-collection-manipulation-with-supergroup-js)

What are you thoughts about this?

billy-addepar:
This version of Ember table is no longer supported. If you want to continue discussion, you can open the issue on [Ember Table Legacy](https://github.com/Addepar/ember-table-legacy)

PR #368, opened 2015-07-31
Example of building treetable with supergroup.js and plain data

artych:
This is an attempt to combine beautiful tree-table (based on financial table example) and local tree-structure constructed by supergroup.js. Inspired by @taras, #356.


taras:
wow, very cool. excited to check this out!

artych:
Year, your feedback would be helpful, and I'd be glad to collaborate on this.

artych:
Help to properly set dependency with `supergroup.js` master branch is appreciated.

This will allow to significantly cleanup the code.  I previously used `bower install` and got stale 1.0.1 version. See Sigfried/supergroup#6 for details. 

I tried to provide commit no and direct link to `https://rawgit.com/Sigfried/supergroup/master/supergroup.js`, but got error "Could not find module lodash" inside `supergroup.js` code.

Issue #369, opened 2015-08-04
Error installing 0.9.1


morten-holm:
Hi 

I'm having trouble installing ember-table 0.9.1:

ember install ember-table@0.9.1
version: 1.13.6
Installed packages for tooling via npm.
installing ember-table
Installing browser packages via Bower...
  not-cached git://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da=git://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da
  cached git://github.com/jquery/jquery-mousewheel.git#3.1.13
  cached git://github.com/components/jqueryui.git#1.11.4
Tag/branch e0d1538cf4f3fd61c5bedd6168df86d651f125da=git://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da does not exist
Error: Tag/branch e0d1538cf4f3fd61c5bedd6168df86d651f125da=git://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da does not exist
...

Ember version: ember -v
version: 1.13.6
node: 0.12.4
npm: 2.13.2
os: darwin x64

azirbel:
Sorry to hear! We've run into this kind of problem before: when you install ember-table, it tries to install https://github.com/Addepar/antiscroll as a Bower dependency. For some reason, it can't find commit `e0d1538cf4f3fd61c5bedd6168df86d651f125da` (even though I can see that it exists at https://github.com/Addepar/antiscroll/commit/e0d1538cf4f3fd61c5bedd6168df86d651f125da).

What is the result of running the following two commands?

```
bower install git://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da

bower install https://github.com/Addepar/antiscroll.git#e0d1538cf4f3fd61c5bedd6168df86d651f125da
```

morten-holm:
Both of the above commands executes just fine - still can't install ember-table though (same error).

PR #370, opened 2015-08-05
Restructured antiscroll dependency

morten-holm:
This fixes the following issue:

https://github.com/Addepar/ember-table/issues/369

My bower version is 1.4.1.


azirbel:
LGTM! Thanks @morten-holm!

azirbel:
Notes-

This is now released as Ember Table v0.9.2.

This change fixes the install process in new versions of Ember CLI (tested with 1.13.7). Unfortunately it will break in old versions of Ember CLI; if installing in an older version, try ember-table v0.9.1.

Since people running `ember install ember-table` are probably new Ember CLI users, and this change keeps ember-table up to date, it makes sense to merge it now.

PR #385, opened 2015-11-25
Fixed findRow, which always returned null, to use findBy().

DanChadwick:
Currently `findRow` always returns null, because the return within the `forEach` closure function only returns from the closure, not from `findRow()`. This PR uses `findBy` to locate the row. It maintains the intended functionality of returning null when the content is not found, rather than undefined as is returned by `findBy`.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -575,13 +575,7 @@ StyleBindingsMixin, ResizeHandlerMixin, {
   },
 
   findRow: function(content) {
-    // TODO(azirbel): Replace with filter
-    this.get('bodyContent').forEach(function(row) {
-      if (row.get('content') === content) {
-        return row;
-      }
-    });
-    return null;
+    return this.get('bodyContent').findBy('content', content) || null;
   },
 
   rowIndex: function(row) {
--- a/tests/unit/ember-table-test.js
+++ b/tests/unit/ember-table-test.js
@@ -1,0 +1,21 @@
+import { test, moduleForComponent } from 'ember-qunit';
+import Ember from "ember";
+
+moduleForComponent('ember-table', '[Unit] TableComponent', {
+  needs: ['template:components/ember-table']
+});
+
+test("findRow returns the row containing the content", function(assert){
+  var rowA = Ember.Object.create({content: 'A'}),
+      rowB = Ember.Object.create({content: 'B'}),
+      rowC = Ember.Object.create({content: 'C'});
+  var table = this.subject({
+    bodyContent: [
+      rowA,
+      rowB,
+      rowC
+    ]
+  });
+
+  assert.equal(table.findRow('B'), rowB, 'The row was found');
+});


addon/components/ember-table.js
```diff
@@ -575,13 +575,8 @@ StyleBindingsMixin, ResizeHandlerMixin, {
...
   findRow: function(content) {
-    // TODO(azirbel): Replace with filter
-    this.get('bodyContent').forEach(function(row) {
-      if (row.get('content') === content) {
-        return row;
-      }
-    });
-    return null;
+    var row = this.get('bodyContent').findBy('content', content);
+    return row ? row : null;
```
derekbrown: Nitpick here, but not sure if we're using the ternary condition in the repo elsewhere. Would rely on @cyril-sf, @jordansmith42, and @embooglement here.

derekbrown:
This is awesome, @DanChadwick. Thanks for contributing.

DanChadwick:
While I think the ternary operator is a good idiom, at least when limited to one or two levels, I do see that this might be more simply put as:
```javascript
return this.get('bodyContent').findBy('content', content) || null;
```
I don't love the above idiom, but it seems to be well established in the javascript world. Potato / Potaaato, I think.

derekbrown:
@DanChadwick Sorry, should've been clearer; I meant just expanding the ternary out into it's `if/else` structure. Either way, thanks again for contributing!

addon/components/ember-table.js
embooglement: Personally I prefer full `if`-`else` statements, although here the statement is simple enough that I don't mind.

DanChadwick:
Swapped out the tertiary `row ? row : null` operator for `|| null`, which is both cleaner and uncontroversial.

cyril-sf:
@DanChadwick Thanks for the contribution! Can you add a test?

DanChadwick:
I have been completely unsuccessful, despite significant effort, in getting testing to run in my Windows environment. I have, however, tested it manually. Since the current implementation is 100% broken (never works), there isn't much risk of committing this. It can't possibly be worse, right? ;)

cyril-sf:
@DanChadwick We're going to help you as the test infrastructure is also not perfect today.

cyril-sf:
@DanChadwick I pushed 2 commits on a branch

https://github.com/Addepar/ember-table/tree/cyril/find-fixrow

One is the test that you should add to your commit, the other commit is not related but useful when debugging/building.

Please squash ba960ef8d5629a92d3d9e3895f606b7f5cd7514a into your commit and we will review/merge it.

DanChadwick:
@cyril-sf - At this point, I suggest that you just merge your `cyril/find-fixrow` branch into your `cyril-ember-2.0` branch. If this `findRow` but is also in `master`, and I don't know that it is, you could backport to master. Then this PR to merge with master can be closed.

billy-addepar:

PR #387, opened 2015-12-02
fix deprecation warning due to using shared getter and setter

ofridagan:
http://emberjs.com/deprecations/v1.x/#toc_computed-properties-with-a-shared-getter-and-setter


billy-addepar:

PR #390, opened 2015-12-04
Fix undefined table component

embooglement:
There is a race condition where if a header cell is removed from the dom during resizing, `tableComponent` becomes undefined in `elementSizeDidChange` because the associated view has already been destroyed by the time the `afterRender` portion of the run loop happens. This throws an exception. This commit properly cancels the scheduled callback when the view is destroyed, preventing the exception.


jordansmith42:
LGTM @cyril-sf any concerns?

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,42 @@
...
+  unit: true,
+  needs: [
+    'template:header-cell'
+  ]
+});
+
+test('The elementSizeDidChange scheduled task is correctly canceled when the element is destroyed', function(assert) {
+  assert.expect(3);
+
+  let headerCell = this.subject({
```
philpee2: I think this can be `const`

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,42 @@
...
+    content: ColumnDefinition.create({}),
+    tableComponent: {
+      columnMode: 'standard'
+    }
+  });
+
+  Ember.run(function() {
```
philpee2: One-liner, bro
`Ember.run(() => headerCell.append());`

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,42 @@
...
+    headerCell.append();
+  });
+
+  assert.ok(!headerCell._scheduledElementSizeDidChange,
```
philpee2: Hm, doesn't seem great that you're testing something that isn't the public API of this view

tests/unit/header-cell-test.js
cyril-sf: `this.append` or `this.render` ( I forgot which version of Ember QUnit is used ). No need to wrap it in a run loop, no?

tests/unit/header-cell-test.js
embooglement: I agree, but I don't know a better way to do it.

tests/unit/header-cell-test.js
embooglement: If you really want me to, I guess.

tests/unit/header-cell-test.js
embooglement: Good catch

cyril-sf:
Does the test fail if you revert the fix?

tests/unit/header-cell-test.js
cyril-sf: It's using the proper API of the library we use, so yes :smile:

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,40 @@
...
+  needs: [
+    'template:header-cell'
+  ]
+});
+
+test('The elementSizeDidChange scheduled task is correctly canceled when the element is destroyed', function(assert) {
+  assert.expect(3);
+
+  const headerCell = this.subject({
+    content: ColumnDefinition.create({}),
```
cyril-sf: can you mock the column definition the same way you do with the `tableComponent`?

tests/unit/header-cell-test.js
embooglement: `this.append` is only defined for tests using `moduleForComponent`, and that only works with actual components, not views. `append` has to be called on the view directly, and must be in a run loop.

cyril-sf: I forgot that this was still a view :disappointed:

tests/unit/header-cell-test.js
embooglement: What do you mean? I only mocked the stuff in `tableComponent` that was necessary for the test to work.

cyril-sf: Do you need to import `ColumnDefinition`?

embooglement: Yes.

embooglement: There's a bunch of initialization stuff it does that's needed, it can't just be a raw object.

embooglement:
Yes, things fail if the fix is changed incorrectly (the test doesn't even make sense if the fix is reverted because then `_scheduledElementSizeDidChange` is never even mentioned, so them failing is meaningless).

cyril-sf:
That's right. I guess not canceling `_scheduledElementSizeDidChange` should be enough to make the test fail.

embooglement:
Yeah, that fails.

cyril-sf:
I realize that `_scheduledElementSizeDidChange` is canceled twice, it feels like we're missing a test now.

embooglement:
How is it cancelled twice?

embooglement:
Oh, like if the function is called without the element being destroyed.

cyril-sf:
It's canceled in `elementSizeDidChange` and in `willDestroyElement`.

cyril-sf:
By looking at the code, canceling in `willDestroyElement` seems actually not needed as it will be done in `elementSizeDidChange` anyway.

embooglement:
I guess I code just always call `elementSizeDidChange` in `willDestroyElement`.

embooglement:
@cyril-sf Added another test and updated `willDestroyElement`. Any other concerns?

embooglement:
@Addepar/fire updated this with comments from @cyril-sf, any other concerns?

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,77 @@
...
+  Ember.run(function() {
+    headerCell.set('width', 100);
+    headerCell.destroy();
+  });
+
+  assert.equal(callCount, 1,
+    'elementSizeDidChange is not called for destroyed elements');
+});
+
+test('The elementSizeDidChange scheduled task is canceled when the task is done', function(assert) {
```
cyril-sf: `is called only once` instead of `is canceled`?

tests/unit/header-cell-test.js
```diff
@@ -0,0 +1,77 @@
...
+    oldElementSizeDidChange.apply(this, arguments);
+  };
+
+  Ember.run(() => headerCell.append());
+
+  assert.equal(callCount, 1,
+    'elementSizeDidChange is called for when initializing elements');
+
+  Ember.run(function() {
+    headerCell.set('width', 100);
```
cyril-sf: Set the width twice to make sure that `scheduleOnce` is used

embooglement: Done

embooglement:
@cyril-sf anything else?

cyril-sf:
LGTM

PR #391, opened 2015-12-09
[WIP] Refactor Column Resizing

embooglement:
Making this logic a little cleaner.


addon/components/ember-table.js
```diff
@@ -315,37 +315,59 @@ StyleBindingsMixin, ResizeHandlerMixin, {
   // they can be. Note that this may fail to arrive at the table width if the
   // resizable columns are all restricted by min/max widths.
   doForceFillColumns: function() {
+    console.log('doForceFillColumns');
```
jordansmith42: ?

addon/components/ember-table.js
```diff
@@ -315,37 +315,59 @@ StyleBindingsMixin, ResizeHandlerMixin, {
...
+    /**
+     * Resizes a column, and returns whether or not the column is now at it's
+     * minimum or maximum.
+     * @param {ColumnDefinition} column The column to be resized
+     * @param {number} totalResizableWidth The total width of the table that
+     *   can be resized
+     * @param {number} availableWidth The total width available in the table
+     * @returns {boolean}
+     */
+    var resizeColumn = function(column, totalResizableWidth, availableWidth) {
```
jordansmith42: Why not pull this out of the `doForceFillColumns` method entirely?

addon/components/ember-table.js
```diff
@@ -315,37 +315,59 @@ StyleBindingsMixin, ResizeHandlerMixin, {
...
+      var newWidth = Math.floor(column.get('width') / totalResizableWidth *
+        availableWidth);
+
+      if (newWidth < column.get('minWidth')) {
```
jordansmith42: You could just cache these rather than getting them multiple times

addon/components/ember-table.js
embooglement: Whoops, forgot that.

addon/components/ember-table.js
embooglement: I was planning on it.

addon/components/ember-table.js
embooglement: Will do.

embooglement:
@jordansmith42 updated that stuff. Any other issues you ssee?

jordansmith42:
LGTM

DanChadwick:
I haven't tried the code, but from reading it, I have some comments:
1. In `_resizeColumn`, why is `availableWidth` updated but the updated value is not used or returned. It needs to be returned unless you intended to change the original algorithm.
2. I have not seen Ember code wordwrapped to 80 columns (just the docblocks). Is there a whitespace style that I'm not aware of? It certainly makes the code harder to read.
3. In `_resizeColumn`, it would be slightly clearer to write `new = width * (resizeable / available)` -- in other words prorating the width. It also avoids any question of operator associativity.
4. A `do - while` loop would avoid having the set the semaphore before the loop.
5. Would the `var`s with the loops would be better as `let`s?
6. I keep wondering if an infinite loop is possible where it tried to give the same space to the same column on each pass. Not sure. 
7. It doesn't seem guaranteed to use all available width due to the use of `floor`.

Again, I didn't try it, but rather just looked at the code.

embooglement:
@DanChadwick I was mostly trying to refactor things to be a little clearer, but keep the existing behavior. Addressing specific comments: 
1) Whoops, pretty bad mistake on my part, will fix.
2) Force of habit on my part.
3) I think that shortening the variable names makes the code harder to understand.
4) True. Might update that.
5) es6 isn't currently set up correctly for ember-table, and I had trouble with `let` and `const`, which I'm going to fix soon.
6) It probably is, but I wanted to keep the original behavior, and address problems with it later
7) Same as 6

embooglement:
Thanks for the feedback though, it's definitely helpful!

DanChadwick:
Re 3) -- Sorry to not be clearer. I wasn't saying that you should shorten the names. I was suggesting a change of the order and adding the parenthesis. There is no functional difference but the suggestion is mentally clearer (at least to me): Width * Fraction that I want to scale it by.

embooglement:
Ah, yeah, I agree. Did that in my new PR (https://github.com/Addepar/ember-table/pull/392)

PR #392, opened 2015-12-10
Continue refactoring doForceFillColumns

embooglement:
Addressing some mistakes I made in my earlier refactor, as well as addressing comments by @DanChadwick.


DanChadwick:
I didn't mean to suggest that you should use fractional widths (by removing floor), but rather in calculating the width of each column, I think the "flooring" to an integer results can result in the table being a few pixels too big or small (I think). 

Once solution would be that when you are adding space (resized wider), use ceil so that you round fractional pixels up, and floor when removing space (resized narrower). I haven't tested this -- its just an idea.

But then, I'm not totally sold on this algorithm because I'm not sure that it terminates under all input conditions.

All this from reading. I haven't tried anything....

Thanks for working on this!

embooglement:
Everything about this algorithm makes me sad. It think it's way more complicated than it needs to be, and really confusing. Long term goal is to rewrite this entirely (I'm not convinced that this iterative approach is the best way to do this, I think there's a mathematical approach that could be simpler).

PR #397, opened 2016-02-19
Add option to allow table height to grow without specified container height

LydiaBarnes:
Currently table content height is calculated to never go higher than its container height so that it can scroll within it. But sometimes a table that just grows infinitely without having to specify a height on the container is desired.
This change adds a flag passed in through the component to tell it whether to restrict to the container height or not.

Table without container height set and flag = true:
screen shot 2016-02-18 at 7 00 13 pm

Table without container height set and flag = false:
screen shot 2016-02-18 at 6 58 34 pm

@Addepar/pants @Addepar/fire

PR #399, opened 2016-02-29
Add option to allow table height to grow without specified container height

LydiaBarnes:
Note: This is a copy of my old PR on the addepar branch https://github.com/Addepar/ember-table/pull/397

Currently table content height is calculated to never go higher than its container height so that it can scroll within it. But sometimes a table that just grows infinitely without having to specify a height on the container is desired.
This change adds a flag passed in through the component to tell it whether to restrict to the container height or not.

Table without container height set and flag = true:
screen shot 2016-02-18 at 7 00 13 pm

Table without container height set and flag = false:
screen shot 2016-02-18 at 6 58 34 pm

@Addepar/pants @Addepar/fire


LydiaBarnes:
Now that hackathon is over... re-tagging
@Addepar/pants @Addepar/fire

embooglement:
This should probably have some tests.

LydiaBarnes:
@Addepar/pants @Addepar/fire 
I added acceptance tests for this, which meant I had to add a bunch of other stuff like an actual route with the table to test. @jordansmith42 and I agreed this is the best way to test my change, since it needs a rendered table to evaluate the heights in the DOM as well as have a parent container (outside of the component) to test the height logic.

I also updated the documentation page that I missed before.

tests/dummy/app/templates/table-height.hbs
```diff
@@ -0,0 +1,31 @@
...
+          {{!// END-SNIPPET}}
+        </div>
+      </div>
+    </div>
+
+    <div class="col-md-12 bumper-30">
+      <h3>Template</h3>
+      {{code-snippet name="table-height-table.hbs"}}
+    </div>
+  </div>
```
LydiaBarnes: I wasn't sure if I should include the controller snippet, since its the same as simple table and has nothing to do with the table height option. (Will work on any table no matter the controller)

tests/acceptance/table-height-test.js
```diff
@@ -0,0 +1,57 @@
...
+    Ember.run(application, 'destroy');
+  }
+});
+
+/**
+ * Gets the height of an element set by CSS (instead of the rendered height)
+ * and strips the 'px'.  This alternative to using .height() is needed because
+ * the literal rendered height will be slightly off from the height we set
+ * through inline CSS, which can cause inaccurate calculations.
+ * @param  {object} element: The DOM element we want to target
```
embooglement: Should be `@param {jQuery}`.

tests/acceptance/table-height-test.js
```diff
@@ -0,0 +1,57 @@
...
+    const firstRow = ['Thu Jul 14 2005', '22.49', '-18.01', '-47.10', '-0.30'];
+    assert.deepEqual(rowText(0), headers, 'headers');
+    assert.deepEqual(rowText(1), firstRow, 'values in first row');
+  });
+});
+
+test('Table height renders properly without a container height set', function(assert) {
+  visit('/table-height');
+  andThen(function() {
+    var tableBodyHeight = getCSSHeight($('.ember-table-body-container'));
```
embooglement: All these `$` should probably be `find`, which could also just be done `getCSSHeight`

tests/dummy/app/controllers/table-height.js
```diff
@@ -0,0 +1,64 @@
+// BEGIN-SNIPPET table-height-controller
```
embooglement: What does this do?

LydiaBarnes: Its for the fancy code snippets in the demo page (like in http://opensource.addepar.com/ember-table/#/simple)

DanChadwick:
@ embooglement: May I ask why `this.$('xxx')` is inferior to `this.$().find('xxx')`? Not disputing, just curious?

tests/dummy/app/controllers/table-height.js
embooglement: Oh. Neat.

embooglement:
I'm not advocating `this.$().find('xxx')`, I'm saying to use the `find` test helper over a raw jQuery call. See: http://emberjs.com/api/classes/Ember.Test.html#method_find

The reason for this is that the `find` helper will scope itself to the actual app, so you'll never accidentally select something outside the application itself (see the definition of `find` at https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/helpers.js#L290).

embooglement:
While it's unlikely that there are any elements in the test runner with a class of `ember-table-body-container`, I think it's a good habit to get into.

DanChadwick:
Ah. My confusion re `$().find`. Thanks for unconfusing me.

tests/acceptance/table-height-test.js
LydiaBarnes: I forgot that `find` is a default ember helper, will update (and put in the function).

embooglement:
No problem.

tests/acceptance/table-height-test.js
```diff
@@ -0,0 +1,57 @@
...
+  } from 'qunit';
+import startApp from '../helpers/start-app';
+import {setRandomSeed} from 'dummy/utils/random';
+
+var application;
+
+module('Acceptance: Table Height', {
+  beforeEach: function() {
+    application = startApp();
+    setRandomSeed(6);
```
billy-addepar: Why do you need to set random seed here? Test result should be deterministic

tests/dummy/app/controllers/table-height.js
```diff
@@ -0,0 +1,64 @@
...
+import Ember from 'ember';
+import ColumnDefinition from 'ember-table/models/column-definition';
+import {randomNumber, randomDate} from '../utils/random';
```
billy-addepar: Dont use `random` in test

billy-addepar:
Use specific number instead of `random` in test. Test should be deterministic and we want to make sure it's reproducible after it fails.

tests/acceptance/table-height-test.js
LydiaBarnes: TBH I copied it from the others, makes sense it wouldn't be needed for this

LydiaBarnes:
@billy-addepar I understand why we shouldn't use random in tests. I'm not really sure how to address this though, other than just repeating all of the rows with the same content. The specific test I wrote doesn't actually test the validity of any of the table content, just the height of the whole thing (which I don't think I can really do with a unit test due to the nature of the way the table renders). 
The table in the controller I created with the random content is the demo page table for the property, which the acceptance tests also uses (and I pretty much straight up copied it). This is how all of the other acceptance tests are set up.  I know "doing it the way its done now" is not always the best excuse, so if using random numbers to set up a demo table is a problem, then we should address that.

billy-addepar:
Ah, it's fine to have it in the dummy app. Just dont use in test though. 

LGTM after the randomSeed is removed

embooglement:
If a seed is explicitly set to a constant in a test, that test is
deterministic. I still am not a big fan of random content though.

tests/acceptance/table-height-test.js
```diff
@@ -0,0 +1,45 @@
...
+  module,
+  test
+  } from 'qunit';
+import startApp from '../helpers/start-app';
+import {setRandomSeed} from 'dummy/utils/random';
+
+var application;
+
+module('Acceptance: Table Height', {
+  beforeEach: function() {
```
embooglement: You should call `setRandomSeed()` here.

embooglement:
After the call to `setRandomSeed` is added, I would say this LGTM.

addon/components/ember-table.js
```diff
@@ -75,6 +75,13 @@ StyleBindingsMixin, ResizeHandlerMixin, {
   // through ctrl/cmd-click or shift-click).
   selectionMode: 'single',
 
+  // A flag to control whether the height of the container is used to
```
embooglement: This should be a multi-line documentation comment. JSDoc only supports comments of exactly the form:
```
/** 
comment text
*/
```

cyril-sf:
I'm going to remove all those tests. If you want to update the documentation, you need to modify the `demo-app` branch.

LydiaBarnes:
@cyril-sf I updated this old branch after your changes, does everything look correct to you?

embooglement:
@LydiaBarnes what's the status of this PR? Are we doing it? If not, can we close it?

LydiaBarnes:
I would like to merge it as it is still useful. I just held off on it, with all the ember-cli updates it was getting confusing. I just want to know if it's okay now (I removed all the old stuff).

angelagu:
Is this still needed?

PR #406, opened 2016-05-26
Wrap callbacks from external libraries inside the run loop

cyril-sf:
This allows to catch with `Ember.error` any raised exceptions.


sreenath-are:
lgtm

PR #410, opened 2016-09-27
Remove problematic CSS !importants

LydiaBarnes:
While refactoring the CSS for tables in the product, I realized there are some `!important`s in Ember Table itself that could be removed. I only addressed the couple that were directly affecting our overwriting CSS, I didn't do a major overhaul. The ones I removed weren't even really needed in the first place, the selectors were specific enough already so I have no idea why !important was used.

@Addepar/design @cyril-sf @shiller-addepar @embooglement

### DIFF
--- a/addon/styles/themes/financial.less
+++ b/addon/styles/themes/financial.less
@@ -54,7 +54,7 @@
   }
 
   .ember-table-left-table-block .ember-table-cell {
-    border-right: 16px solid transparent !important;
+    border-right: 16px solid transparent;
   }
 
   /* Header Container */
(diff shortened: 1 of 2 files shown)


LydiaBarnes:
BTW I also tested on the demo-app branch.

addon/styles/themes/financial.less
```diff
@@ -54,7 +54,7 @@
   }
 
   .ember-table-left-table-block .ember-table-cell {
-    border-right: 16px solid transparent !important;
```
raemadeline: finally! ❤️

LydiaBarnes:
retest

PR #422, opened 2017-09-15
Rename table

billy-addepar:
Rename `ember-table-2` -> `ember-table`


billy-addepar:
@pzuraq

pzuraq:
Let's not disable travis and fix the tests now, we should be running them on all versions

billy-addepar:
Given that it takes time to fix the `ember-try` and publish Addepar stylegiude, I think we should go ahead and merge this PR.

pzuraq:
I'm ok with it if we remove the changes to Ember try config, if tests are going to fail anyways we shouldn't mess with the test suite.

LICENSE.md
```diff
@@ -1,9 +1,25 @@
-The MIT License (MIT)
+Copyright © 2017 Addepar, Inc. All Rights Reserved.
```
billy-addepar: This license is the same as the old one:

https://github.com/Addepar/ember-table/blob/addepar/LICENSE.md

Only the year is changed.

billy-addepar:
Yay, tests pass. I don't why it passes but it passes. Ready for review again

.travis.yml
```diff
@@ -13,11 +13,6 @@ cache:
 env:
   # we recommend new addons test the current and previous LTS
   # as well as latest stable release (bonus points to beta/canary)
-  - EMBER_TRY_SCENARIO=ember-lts-2.8
```
pzuraq: We still need to revert these changes, and we really should add `ember-1.11` too but we can get this merged for now

billy-addepar:
@pzuraq alright, I added a number of test scenarios in the travis file, including 1.13, 2.8, etc. All of them pass now. 1.11 is failing `You can only pass attributes (such as name=value) not bare values to a helper for a View found in` and I have not figure out how to fix that in test. But at least we have coverage for a number of Ember versions.

pzuraq:
Ah, to fix that just add `import 'vertical-collection/-debug';` to `test-helper.js`

billy-addepar:
Adding `import 'vertical-collection/-debug';` does not solve it. I still get the same issue

billy-addepar:
Yay, all tests pass including 1.11

PR #426, opened 2017-09-28
Table theme

billy-addepar:
Allow table to have custom theme.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -19,7 +19,7 @@ const SELECTION_MODE_MULTIPLE = 'multiple';
 
 export default class EmberTable2 extends Component {
   @property attributeBindings = ['style:style'];
-  @property classNames = ['et-table'];
+  @property classNameBindings = ['tableClass'];
   @property tagName = 'table';
 
   @property layout = layout;
@@ -55,6 +55,11 @@ export default class EmberTable2 extends Component {
    * (multiple rows can be selected through ctrl/cmd-click or shift-click).
    */
   @property selectionMode = 'single';
+
+  /**
+   * Default theme of this table.
+   */
+  @property tableTheme = 'et-default-theme';
 
   /**
    * A temporary element created when moving column. This element represents the current position
@@ -104,6 +109,12 @@ export default class EmberTable2 extends Component {
   @property _width = 0;
 
   @property lastSelectedIndex = -1;
+
+  @computed('tableTheme')
+  tableClass() {
+    const tableTheme = this.get('tableTheme');
+    return `et-table ${tableTheme}`;
+  }
 
   @computed('numFixedColumns')
   get hasFixedColumn() {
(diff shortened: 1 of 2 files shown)


billy-addepar:
@Addepar/ice

pzuraq:
I think it's probably better to just allow users to implement their own styles using the provided class names

billy-addepar:
but they would have to override many of our default theme as well. That's what I want to avoid. If I want to have custom theme, I would create it by my own. Otherwise, I would use the default one

pzuraq:
In general, addon/styles should only include the bare minimum of CSS required for the addon to function as there is no way for the user to exclude the styles from the build. If a user doesn't want our default styles, they probably also don't want to ship it to all of their users.

We can either include the styles in app/styles where they can be optionally imported, or better yet style it internally. We'll likely be creating a second addon to extract the table components we use, so that's the most natural place to add those styles.

billy-addepar:
Done. 

I would rather move the style to `app.scss` than creating a new addon for it. Having an addon just for the style of this table is overkill.

addon/components/ember-table.js
```diff
@@ -19,7 +19,7 @@ const SELECTION_MODE_MULTIPLE = 'multiple';
 
 export default class EmberTable2 extends Component {
   @property attributeBindings = ['style:style'];
-  @property classNames = ['et-table'];
+  @property classNameBindings = ['tableClass'];
```
pzuraq: Revert this, we should just allow users to style it themselves, no need to add custom class names

addon/components/ember-table.js
```diff
@@ -57,6 +57,11 @@ export default class EmberTable2 extends Component {
   @property selectionMode = 'single';
 
   /**
+   * Theme of this table. Override this property to have custom theme for the table.
+   */
+  @property tableTheme = 'et-default-theme';
```
pzuraq: Not needed

addon/components/ember-table.js
```diff
@@ -105,6 +110,12 @@ export default class EmberTable2 extends Component {
 
   @property lastSelectedIndex = -1;
 
+  @computed('tableTheme')
```
pzuraq: Not needed

tests/dummy/app/styles/app.scss
```diff
@@ -26,3 +26,70 @@
 .custom-row {
   background: green;
 }
+
```
pzuraq: This should all be moved to `/app/styles/ember-table/default.scss`, not in the Dummy app. Check out how `ice-pop` and other internal addons work, this allows users to import the styles optionally.

billy-addepar:
Done

billy-addepar:
Why does it override user's style? I thought for SASS, if you don't import it, it won't be used?

pzuraq:
The contents of the `app` folder get directly merged into the main app. If there is another file named `/app/styles/default.scss` it will be overwritten.

billy-addepar:
Done

PR #428, opened 2017-10-04
Header & row action

billy-addepar:
- Adds support for sending action from table header / row.

### DIFF
--- a/addon/components/ember-table-row.js
+++ b/addon/components/ember-table-row.js
@@ -1,16 +1,18 @@
 import { get } from '@ember/object';
 import Component from '@ember/component';
 import { computed } from 'ember-decorators/object';
+import { classNames } from 'ember-decorators/component';
 import { property } from '../utils/class';
 
 import layout from '../templates/components/ember-table-row';
 import { A as emberA } from '@ember/array';
 import { readOnly } from '@ember/object/computed';
 
+@classNames('et-tr')
 export default class EmberTableRow extends Component {
   @property layout = layout;
   @property tagName = 'tr';
-  @property classNames = ['et-tr'];
+
   @property _cells = null;
   @property classNameBindings = ['isSelected']
   /**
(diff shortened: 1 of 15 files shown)


billy-addepar:
@pzuraq

addon/components/ember-table-row.js
```diff
@@ -75,6 +76,13 @@ export default class EmberTableRow extends Component {
   }
 
   click(event) {
-    this.get('targetObject').send('onRowClicked', event, this.get('rowIndex'));
+    const tableObject = this.get('row.api.targetObject');
+    tableObject.send('onRowClicked', event, this.get('rowIndex'));
+
+    this.sendAction('onRowClicked', event, this.get('rowIndex'), this.get('rowValue'));
```
billy-addepar: There are 2 actions sent in this case: one for table for row selection. The other is for outer controller

pzuraq: I think `onRowClicked` here is a bit verbose, from the user standpoint they already know they're passing in actions to the row. Let's do `onClick` instead

addon/components/ember-table-row.js
```diff
@@ -75,6 +76,13 @@ export default class EmberTableRow extends Component {
...
+  }
+
+  doubleClick(event) {
+    this.sendAction('onRowDoubleClicked', event, this.get('rowIndex'), this.get('rowValue'));
```
pzuraq: same here, `onDoubleClick`

addon/components/ember-table-row.js
```diff
@@ -77,12 +77,12 @@ export default class EmberTableRow extends Component {
 
   click(event) {
     const tableObject = this.get('row.api.targetObject');
-    tableObject.send('onRowClicked', event, this.get('rowIndex'));
+    tableObject.send('onClick', event, this.get('rowIndex'));
```
pzuraq: This is different because we are not sending this action externally. Really, if the table were using the row component directly, it would look like:

```
{{#ember-table as |row|}}
  {{ember-table-row onClick="onRowClicked"}}
{{/ember-table}}
```
So this action and the action on the table should remain `onRowClicked`

PR #429, opened 2017-10-04
Update layout upon change in number of columns

dmoon1221:
Same changes as https://github.com/Addepar/ember-table-legacy/pull/5 but also adding here in order to get the bug fix into the current candidate, which still depends on `ember-table`.


dmoon1221:
@Addepar/ice

PR #430, opened 2017-10-06
Prepare alpha

billy-addepar:
- Update README
- Update dummy app home page to link to `simple` & `demo` route
- Fix a bug caused by  aa70543c25220044ba7fd4c843dec681168e173b

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,19 @@ ember install ember-table
 
 To use `Ember Table`, you need to create `columns` and `rows` dataset.
 
-`columns` is an array of `ColumnDefinition` which has multiple fields to define behavior of column.
-Two required field for `ColumnDefinition` is `columnName` and `valuePath`.
+`columns` is an array of objects which has multiple fields to define behavior of column.
+Two required field in each object is `columnName` and `valuePath`.
 
 ```javascript
   columns: [
-    ColumnDefinition.create({
+    {
       columnName: `Open time`,
       valuePath: `open`
-    }),
-    ColumnDefinition.create({
+    },
+    {
       columnName: `Close time`,
       valuePath: `close`
-    })
+    }
   ]
 ```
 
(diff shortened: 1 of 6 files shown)


billy-addepar:
@pzuraq

README.md
```diff
@@ -1,26 +1,108 @@
...
 
-* `npm test` (Runs `ember try:each` to test your addon against multiple Ember versions)
-* `ember test`
-* `ember test --server`
+`columns` is an array of `ColumnDefinition` which has multiple fields to define behavior of column.
+Two required field for `ColumnDefinition` is `columnName` and `valuePath`.
 
-## Building
+```javascript
+  columns: computed(function() {
```
pzuraq: Let's show this as just an array of POJOs no need to have a computed or push into it.

README.md
```diff
@@ -1,26 +1,108 @@
...
+
+### Template
+
+The following template renders a simple table.
+
+```
+  {{#ember-table
+    rows=rows
+    columns=columns
+    as |t|
```
pzuraq: Let's do as |row| or as |r| here

billy-addepar:
Done

README.md
```diff
@@ -1,26 +1,104 @@
...
-* `npm test` (Runs `ember try:each` to test your addon against multiple Ember versions)
-* `ember test`
-* `ember test --server`
+`columns` is an array of `ColumnDefinition` which has multiple fields to define behavior of column.
+Two required field for `ColumnDefinition` is `columnName` and `valuePath`.
 
-## Building
+```javascript
+  columns: [
+    ColumnDefinition.create({
```
pzuraq: Just POJOs, we shouldn't enforce the ColumnDefinition class. There's no need for it.

billy-addepar: Right now, it's based on `ColumnDefinition`. We have not found a good solution with using POJO because we want to listen to column width. That's why I keep it for now. We can change it to POJO if needed in the beta version

pzuraq: You can listen to property changes on POJOs, as long as you use `get/set` which is standard in Ember. The only thing you can't listen to directly in 1.11 are changes to array length, which can be solved by re-setting the array after changes.

billy-addepar:
Done

PR #434, opened 2017-10-10
namespace pkg under addepar org

majuscule:
@pzuraq


pzuraq:
See my comment on the Jira issue, we don't want to change the scope, just change NPM permissions. If that's not possible we'll have to figure out a different way to get these published.

PR #436, opened 2017-10-31
Use RAF Scheduler + bug fix

billy-addepar:
- Use RAF Scheduler
- Use `get` to get attribute value instead of using direct object access.


billy-addepar:
@Addepar/ice

Issue #437, opened 2017-11-07
row.collapse is crashing


csukda:
I tried to collapse and expand rows manually and show a "+" or "-" sign when collapsed or not. So I had to access the collapse attribute. I collapsed the rows initially by calling collapseNode. After that it crashes by changing the collapse attribute with "row.collapse =" when calling expand or collapseNode. Exchanging "row.collapse=" with "Ember.set( )" solves this problem.

billy-addepar:
@csukda you're right. It should be using `set` instead of direct property access. 

Can you open a PR to fix that? Also note that we avoid using global Ember name space. Import the `set` instead. Example: https://github.com/Addepar/ember-table/blob/master/addon/components/ember-table.js#L11

PR #438, opened 2017-11-07
Fix issue #437

csukda:
There are other occasions where row.<attribute> is used, please look forward if it's better to use Ember.set instead


csukda:
@billy-addepar 
i merged our fix with yours accidentally, is this a problem for you?

billy-addepar:
@csukda can you do rebase instead?

PR #440, opened 2017-11-07
Fix collapse bug

billy-addepar:
Repro steps: 

1) Collapse Group 2
screen1

2) collapse Header 0
screen2

3) Expand Header 0...

The row order is not in correct order. 

This bug happens because the previous node of `Header 1` is updated twice but we have only 1 variable to store previous node. Correct solution is to have an array to store previous nodes. Previous node could be pushed or popped from array when row is collapsed/expanded.


billy-addepar:
@pzuraq

PR #442, opened 2017-11-13
Pass list of arguments instead of the first arguments.

billy-addepar:
The `onHeaderEvent` API currently passes only 1 argument to the parent component/controller. It should pass the list of all object instead.


billy-addepar:
@pzuraq

PR #444, opened 2017-11-14
Allow custom header 

billy-addepar:
- Rename previous `ember-table-header` to `ember-table-header-cell`

- Add `ember-table-header` to allow custom header.


billy-addepar:
@pzuraq

PR #446, opened 2017-11-16
Support subcolumns

billy-addepar:
- Support subcolumns in the header. As a start, we only support 1 layer of subcolumn.
- The `bodyColumns` refers to the `columns` used in the body for rows. In normal case, it's the same as `columns`. If header has subcolumns, it's the concatenation of all `subcolumns.`

screen shot 2017-11-15 at 4 17 47 pm

### DIFF
--- a/addon/components/ember-table-base-cell.js
+++ b/addon/components/ember-table-base-cell.js
@@ -5,6 +5,7 @@ import { action, computed } from 'ember-decorators/object';
 import { get } from '@ember/object';
 import { property } from '../utils/class';
 import { scheduler, Token } from 'ember-raf-scheduler';
+import { isNone } from '@ember/utils';
 
 export default class EmberTableCell extends Component {
   @property tagName = 'td';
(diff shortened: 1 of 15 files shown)


billy-addepar:
@Addepar/ice

addon/components/ember-table-base-cell.js
```diff
@@ -39,7 +40,14 @@ export default class EmberTableCell extends Component {
...
   scheduleSync() {
     scheduler.schedule('sync', () => {
-      const { height } = this.element.getBoundingClientRect();
+      let { height } = this.element.getBoundingClientRect();
+
+      if (this.get('isFixed')) {
+        // Get the max height of the div element height and other cells height in case other cells
+        // are empty.
+        const fixedDivElement = this.element.getElementsByClassName('is-fixed')[0];
+        height = Math.max(height, fixedDivElement.getBoundingClientRect().height);
```
pzuraq: @billy-addepar what happens if the user changes the row height or contents, and the row should shrink in height?

addon/components/ember-table-base-cell.js
```diff
@@ -39,7 +40,14 @@ export default class EmberTableCell extends Component {
 
   scheduleSync() {
     scheduler.schedule('sync', () => {
-      const { height } = this.element.getBoundingClientRect();
+      let { height } = this.element.getBoundingClientRect();
+
+      if (this.get('isFixed')) {
```
pzuraq: We should revert this commit until we're ready to fully address the issues of keeping the row height and cell height in sync. In the meantime, we can opt into the performance improvement of using fixed heights.

billy-addepar:
@Addepar/ice This is ready for review again.

PR #453, opened 2017-11-22
API to set row height

billy-addepar:


billy-addepar:
@Addepar/ice

pzuraq:
So we should match the vertical-collection API since `estimateHeight` needs to be passed anyways. So we should have `estimateHeight` which, for now, we can assume is a pixel value, and `staticHeight` which is a boolean that turns on using the fixed height

billy-addepar:
In that case, I think we can remove the `estimateHeight` param and make it a computed property. Its value is the same with `rowHeight` if user specifies or 20 by default.

pzuraq:
ah, no actually that's perfect. We should keep `estimateRowHeight`, and user should be able to also pass `staticHeight=true`. If they do, we assume `estimateRowHeight` is the actual row height.

addon/components/ember-table.js
```diff
@@ -128,6 +129,18 @@ export default class EmberTable2 extends Component {
     return Number.isInteger(numFixedColumns) && numFixedColumns !== 0;
   }
 
+  /**
+   * Estimated height for each row.
+   */
+  @computed('staticHeight')
```
pzuraq: We should match the API in vertical-collection. `estimateRowHeight` is a number representing the height. `staticHeight` is a boolean changing the mode of the table.

addon/components/ember-table-row.js
```diff
@@ -75,6 +77,15 @@ export default class EmberTableRow extends Component {
     return this.get('selectedRows').indexOf(this.get('rowValue')) >= 0;
   }
 
+  @computed('row.api.rowHeight')
```
pzuraq: We should pass `estimateRowHeight` and `staticHeight` to the row and only apply the style if `staticHeight === true`

addon/components/ember-table.js
billy-addepar: I don't think we should have 2 variables in this case. One should be enough. User does not need to know about vertical collection implementation details.

PR #454, opened 2017-11-22
chore(collection): update vertical collection

pzuraq:


pzuraq:
@Addepar/ice

addon/templates/components/ember-table.hbs
```diff
@@ -18,20 +18,21 @@
 </thead>
 
   <div class="et-tbody-container" data-test-body-container>
-    {{#vertical-collection rows
-      estimateHeight=estimateRowHeight
-      tagName="tbody"
-      class="et-tbody"
+    <tbody class="et-tbody">
+      {{#vertical-collection rows
```
pzuraq: `vertical-collection` is now tagless, so we need to create an external `tbody` element and add a `containerSelector`

tests/helpers/wait-for-render.js
```diff
@@ -1,11 +0,0 @@
-import { Promise as EmberPromise } from 'rsvp';
-import wait from 'ember-test-helpers/wait';
-
-export default function() {
```
pzuraq: No need to use this waiter anymore, we have the raf scheduler waiter

PR #458, opened 2017-12-01
Table PageObject 

billy-addepar:
- Create initial table/body/header page objects for test. Use page object in table test.
- This PR does *not* support column action on table (column resizing, scrolling) yet. It will be done in subsequence PR.

### DIFF
--- a/addon-test-support/pages/ember-table-body-page.js
+++ b/addon-test-support/pages/ember-table-body-page.js
@@ -1,4 +1,4 @@
-import PageObject, { collection } from 'ember-classy-page-object';
+import PageObject, { collection, hasClass } from 'ember-classy-page-object';
 import { findAll } from 'ember-native-dom-helpers';
 import { findElement } from 'ember-classy-page-object/extend';
 import { getter } from 'ember-cli-page-object/macros';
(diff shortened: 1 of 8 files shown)


billy-addepar:
@Addepar/ice

addon-test-support/pages/ember-table-body-page.js
```diff
@@ -0,0 +1,44 @@
...
+     * Returns the height of selected row.
+     */
+    height: getter(function() {
+      return findElement(this).offsetHeight;
+    }),
+
+    /**
+     * Checks if the selected row contains a specific class.
+     */
+    containsClass(clazz) {
```
pzuraq: I think this may be something that should be added when you `extend` the table, will comment below

addon-test-support/pages/ember-table-body-page.js
```diff
@@ -0,0 +1,44 @@
...
+      return findElement(this).className.indexOf(clazz) >= 0;
```
pzuraq: I think this may be something that should be added when you extend the table, will comment below

addon-test-support/pages/ember-table-body-page.js
```diff
@@ -0,0 +1,44 @@
...
+     * List of all cells for the selected row.
+     */
+    cells: collection({
+      scope: 'td'
+    }),
+
+    /**
+     * Returns the height of selected row.
+     */
+    height: getter(function() {
```
pzuraq: You shouldn't be using `getter`, you can do:

```js

get height() {

}
```

instead. I'm hoping to pull this functionality back upstream to ember-cli-page-object at some point.

tests/integration/components/ember-table-test.js
```diff
@@ -205,12 +226,18 @@ for (let customHeader of customHeaderTests) {
 
 test('Test custom row', async function(assert) {
   await setupFullTable(this, {}, {}, DEFAULT_ROW_COLUMN_COUNT, 'custom-row');
-  assert.ok(find('tbody tr').className.indexOf('custom-row') >= 0, 'Table has custom row');
+
+  let tablePage = TablePage.create();
+
+  assert.ok(tablePage.body.rows.eq(0).containsClass('custom-row'), 'Table has custom row');
```
pzuraq: To do this without a generic `containsClass` method, you would do:

```js
tablePage = TablePage.extend({
  body: {
    rows: collection({
      isCustomRow: hasClass('custom-row')
    })
  }
}).create();
```

In actual usage, this is most likely what people will be doing. If you'd like to create a generic version of the page for these tests, you could also add it in the extension. The point is that in general people will not have generic classes on tables, but specific classes.

pzuraq:
Looks good besides some minor changes

tests/integration/components/ember-table-test.js
billy-addepar: Hmm, it seems the body is overriden instead of being merged. Let's pair tomorrow.

twokul:
@billy-addepar could you squash your commits?

billy-addepar:
This is ready for review again

PR #459, opened 2017-12-02
Add renderAll property

tythewebguy:
Add the `renderAll` property to `ember-table.hbs` and `ember-table.js` so occlusion can be disabled in the `vertical-collection` component. This allows all table content to be rendered in the DOM so it can be printed. By default, lazy rendering with `vertical-collection` only adds a sub-set of table rows to the DOM.

`renderAll=true` renders all table content to the DOM.
`renderAll=false` (the default) uses `vertical-collection` to occlude table data that is not visible.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -129,6 +129,12 @@ export default class EmberTable2 extends Component {
    */
   @property _width = 0;
 
+  /**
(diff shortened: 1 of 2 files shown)


billy-addepar:
@tythewebguy can you add some test here?

PR #462, opened 2017-12-05
Add yarn.lock to gitignore

billy-addepar:


pzuraq:
should also likely remove `yarn.lock` from this PR

PR #464, opened 2017-12-08
Add column resizing & footer to Page Objects in test

billy-addepar:


billy-addepar:
@Addepar/ice

addon/components/ember-table.js
```diff
@@ -175,8 +175,10 @@ export default class EmberTable2 extends Component {
     });
 
     this._tableResizeSensor = new ResizeSensor(this.element, () => {
```
pzuraq: This seems like it should be a separate PR/bugfix. Also, we should try to use `ember-resize-sensor` soon, that's what we built it for after all

PR #465, opened 2017-12-08
Handle 0 length columns gracefully

billy-addepar:


billy-addepar:
@Addepar/ice

addon/components/ember-table.js
```diff
@@ -382,7 +383,10 @@ export default class EmberTable2 extends Component {
   ) horizontalScrollWrapperStyle() {
     let columns = this.get('bodyColumns');
     let visibility = this.get('_width') < this.get('allColumnWidths') ? 'visibility' : 'hidden';
-    let left = this.get('hasFixedColumn') ? get(columns[0], 'width') : 0;
+    let left = 0;
+    if (get(columns, 'length') > 0) {
```
pzuraq: You could probably combine the ternary below here as well:

```
let left;

if (get(columns, 'length') > 0 && this.get('hasFixedColumn')) {
  left = get(columns[0], 'width');
} else {
  left = 0;
}
```

I put the `else` there because to me it feels like a really long ternary - you aren't assigning a value and then mutating it, you are conditionally assigning one value or the other. You could do it either way though.

billy-addepar:
Bump

PR #466, opened 2017-12-11
ES getter function with @computed to avoid DEPRECATION warnings

opendba:
Simple change in .js files to avoid deprecation warnings from decorators.js

deprecate.js:115 DEPRECATION: using @computed with functions directly will be removed in future versions, using ES getter/setter functions instead [deprecation id: macro-computed-deprecated]


billy-addepar:
👍

billy-addepar:
Not sure why lint check fails on test. Let me check.

billy-addepar:
Tests pass now. There is nothing wrong with your PR. Travis has an issue caching wrong dependency. Clearing Travis fixes it.

PR #467, opened 2017-12-15
Put resize sensor code in a run loop as resize sensor uses RAF

billy-addepar:
Under the hood, the `ResizeSensor` is calling the callback in a RAF. This breaks tests when table size changes. To fix this, we put the callback in a `run` loop. 

Further work to use `ember-resize-sensor` to replace ResizeSensor lib comes in separate PR.


billy-addepar:
@Addepar/ice

PR #468, opened 2017-12-18
Enable custom footer event

billy-addepar:


tests/integration/components/ember-table-test.js
```diff
@@ -314,3 +320,68 @@ test('Custom footer', async function(assert) {
...
+            row=r
+            as |cell|
+          }}
+            {{cell.value}}
+          {{/ember-table-row}}
+        {{/ember-table}}
+      </div>
+    `);
+  } else {
+    this.on('onTableFooterEvent', callback);
```
pzuraq: This style should work all the way through modern Ember versions, we use this in vertical-collection

tests/integration/components/ember-table-test.js
```diff
@@ -314,3 +320,68 @@ test('Custom footer', async function(assert) {
...
+
+  let columnCount = 10;
+  let columns = generateColumns(columnCount);
+  let footerRows = generateCustomFooterRows(columns);
+
+  this.set('columns', columns);
+  this.set('rows', generateRows(10, columnCount));
+  this.set('footerRows', footerRows);
+
+  if (Ember.VERSION >= '1.13') { // eslint-disable-line no-restricted-globals, no-undef
```
pzuraq: `Ember.VERSION` is a string. This comparison works right now because `1.13` was the last of the 1.* cycle, so all versions in 2.* _are_ bigger, string-wise. It would fail when `2.13 >= 2.8` for instance. So, we should always use a semver library, or `ember-compatibility-helpers`

PR #469, opened 2017-12-19
Send swapping columns to outer controller.

billy-addepar:
This might be useful to let controller know when 2 columns have been swapped as in case controller wants to load new data or update table view.


billy-addepar:
@Addepar/ice

billy-addepar:
The name `onColumnReordered` is actually better and avoid confusion.

PR #470, opened 2017-12-19
Use RAF Scheduler

billy-addepar:
Replace calling RAF with using RAF scheduler so that it could be canceled when component is destroyed. 

This also adds destroy state of component inside `ResizeSensor` callback.  Under the hood, it is using RAF but since we cannot replace it, we will add destroy state checking before executing the callback.


billy-addepar:
@Addepar/ice

cyril-sf:
Can we test that behavior?

billy-addepar:
Not sure how we can test that. There is no functionality change

billy-addepar:
@Addepar/ice  Test added

PR #476, opened 2018-01-09
Fix table collapse bug

billy-addepar:


billy-addepar:
@pzuraq

pzuraq:
Can you squash these commits and add a descriptive commit message?

billy-addepar:
Done

PR #484, opened 2018-02-02
feat(checkbox): Adds a checkbox option to columns

pzuraq:
Adds a checkbox option to table columns which can be used to toggle the row.
Also refactors row selection to be DDAU, and adds tests and creates a simpler
generateTable function

### DIFF
--- a/addon/components/-private/simple-checkbox.js
+++ b/addon/components/-private/simple-checkbox.js
@@ -1,0 +1,58 @@
+import Component from '@ember/component';
+import { tagName, attribute } from 'ember-decorators/component';
+import { argument } from '@ember-decorators/argument';
+import { type, optional } from '@ember-decorators/argument/type';
+import { Action } from '@ember-decorators/argument/types';
+
+@tagName('input')
+export default class SimpleCheckbox extends Component {
+  value = null;
+
+  @attribute
+  type = 'checkbox';
+
+  @argument({ defaultIfUndefined: true })
+  @type('boolean')
+  @attribute
+  checked = false;
+
+  @argument({ defaultIfUndefined: true })
+  @type('boolean')
+  @attribute
+  disabled = false;
+
+  @argument({ defaultIfUndefined: true })
+  @type('boolean')
+  @attribute
+  indeterminate = false;
+
+  @argument
+  @type('any')
+  @attribute
+  value = null;
+
+  @argument
+  @type(optional(Action))
+  onClick = null;
+
+  @argument
+  @type(optional(Action))
+  onChange = null;
+
+  click(event) {
+    this.sendAction('onClick', event);
+  }
+
+  change(event) {
+    let checked = this.element.checked;
+    let indeterminate = this.element.indeterminate;
+    let value = this.get('value');
+
+    // Checked and indeterminate state have been changed, but that's not DDAU!
+    // Reset the change, send the action and wait for it to be changed manually
+    this.element.checked = this.get('checked');
+    this.element.indeterminate = this.get('indeterminate');
+
+    this.sendAction('onChange', checked, { value, indeterminate }, event);
+  }
+}
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -1,4 +1,6 @@
 /* global ResizeSensor */
+import { assert } from '@ember/debug';
+
 import { action, computed } from 'ember-decorators/object';
 import { tagName, classNames } from 'ember-decorators/component';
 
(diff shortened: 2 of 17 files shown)


tests/helpers/generate-table.js
```diff
@@ -0,0 +1,135 @@
+import hbs from 'htmlbars-inline-precompile';
+import { A as emberA } from '@ember/array';
+import wait from 'ember-test-helpers/wait';
+
+const alpha = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
+
+function hex(a) {
```
billy-addepar: This function looks complicated. If I am not wrong, you are trying to convert a number to Excel style alphabet  index. Can you use simpler function in tests?

tests/helpers/generate-table.js
```diff
@@ -0,0 +1,135 @@
...
+  let s = '';
+  for (let i = 0; i < c; i++) {
+    s = alpha.charAt(a % 26) + s;
+    a = Math.floor(a / 26);
+  }
+
+  return s;
+}
+
+const fullTable = hbs`
```
billy-addepar: There are many duplication of this helper file and the ones in `test-scenario`. Can you merge them?

billy-addepar:
General approach looks good. Just need some clean up

addon/components/-private/simple-checkbox.js
```diff
@@ -0,0 +1,49 @@
...
+import { type } from '@ember-decorators/argument/type';
+
+@tagName('input')
+export default class SimpleCheckbox extends Component {
+  value = null;
+
+  @attribute
+  type = 'checkbox';
+
+  @argument({ defaultIfUndefined: true })
```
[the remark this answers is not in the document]
pzuraq: Yeah, I think so

addon/components/ember-table.js
```diff
@@ -509,10 +523,57 @@ export default class EmberTable extends Component {
...
       selectedRows: this.selectedRows,
-      staticRowHeight
+      staticRowHeight,
+
+      selectRow: this.selectRow
     };
   }
 
+  selectRow = (rowIndex, { toggle, range }) => {
+    let rows = this.get('rows');
```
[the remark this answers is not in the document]
pzuraq: check tests, SELECTION_MODE_NONE is simply not adding an action for onSelect

tests/helpers/generate-table.js
pzuraq: I copied this from online somewhere, I’m sure there’s a lib somewhere. The point is it’s converting us to base26 instead of just modulo, so each cell is verifiably unique.

tests/helpers/generate-table.js
pzuraq: That’s the plan! The APIs are different so I didn’t want to block this PR on them, the next one will be cleanup and expand tests in general, then Tree model refactors

tests/helpers/generate-table.js
billy-addepar: Can you use simpler method here? Like using "Cell" + the input number itself if you want to identify each cell.

This function takes me a while to figure what it does. If there is a bug, it's hard to debug. This is for testing purpose and we do not really need complicated function.

tests/helpers/generate-table.js
billy-addepar: Sounds good to me

billy-addepar:
I tried out this branch and got an error:

```
"Error: Assertion Failed: Attempted to assign 'isSelected' value on a (unknown mixin) component, but no argument was defined for that key. Use the @argument helper on the class field to define an argument which can be passed into the component
 at new EmberError (http://localhost:4200/assets/vendor.js:42116:25)↵    at Object.assert (http://localhost:4200/assets/vendor.js:42359:15)↵    at Function.create (http://localhost:4200/assets/vendor.js:87979:49)↵    at Function.superWrapper [as create] (http://localhost:4200/assets/vendor.js:74175:22)↵    at FactoryManager.create (http://localhost:4200/assets/vendor.js:38866:25)↵    at Proxy.create (http://localhost:4200/assets/vendor.js:38610:20)↵    at CurlyComponentManager.create (http://localhost:4200/assets/vendor.js:43607:31)↵    at http://localhost:4200/assets/vendor.js:31053:33↵    at AppendOpcodes.evaluate (http://localhost:4200/assets/vendor.js:29452:13)↵    at VM.next (http://localhost:4200/assets/vendor.js:36628:32)"

```

tests/helpers/generate-table.js
pzuraq: we need to identify each cell by row and column, which is a change I made to the testing template as well. You can't just check that a cell has content `A`, it need to be `1A` or `2A` or `3A` or `1AA`, so on. 

Alternatively, we can try to have some longer content such as `Row1Column1` to check against, this just seemed simpler (I just googled base26 and copied stack overflow). We can also try to see if there's a Node library that does this, it should be very standard.

pzuraq: https://www.npmjs.com/package/base26

pzuraq: Looks like there isn't an AMD build yet so unfortunately we can't use it 😕 I really don't think it's that big a deal, it's a simple function unlikely to have bugs.

billy-addepar: Cool to me. I would not block the PR because of this.

PR #485, opened 2018-02-05
chore(deps): Update legacy shim and @argument

pzuraq:
After some recent research we've been able to make the legacy class transforms and `@argument` more efficient overall, and more inline with the future version of ES classes. This PR bumps the dependencies to the most recent releases.

### DIFF
--- a/addon/components/ember-table-footer.js
+++ b/addon/components/ember-table-footer.js
@@ -30,7 +30,7 @@ export default class EmberTableFooter extends EmberTableBaseCell {
   }
 
   @action
-  onFooterEvent() {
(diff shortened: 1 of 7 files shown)


addon/components/ember-table-footer.js
```diff
@@ -30,7 +30,7 @@ export default class EmberTableFooter extends EmberTableBaseCell {
   }
 
   @action
-  onFooterEvent() {
+  sendFooterEvent() {
     this.sendAction('onFooterEvent', ...arguments);
```
billy-addepar: Do you want to update this name too for consistency?

PR #486, opened 2018-02-07
[tests] Cleanup tests

pzuraq:
This PR cleans up the tests by utilizing page objects and breaking
out tests into various files. It also attempts to start formalizing
some scenario helpers that can be used to run the same modules against
multiple scenarios.


pzuraq:
@Addepar/ice

tests/dummy/app/components/custom-footer.js
```diff
@@ -10,6 +13,14 @@ export default class CustomFooter extends Component {
   @argument rowIndex;
   @argument onFooterEvent;
 
+  @computed('column.valuePath', 'rowValue')
+  get footerValue() {
```
billy-addepar: why do you need this footerValue for this custom component?

tests/integration/components/basic-test.js
```diff
@@ -0,0 +1,103 @@
...
+    test('occlusion works', async function(assert) {
+      let rowCount = 100;
+      let columnCount = 10;
+
+      await generateTable(this, { columnCount, rowCount });
+
+      assert.ok(table.rows.length < rowCount, 'not all rows have been rendered');
+      assert.equal(table.getCell(0, 0).text.trim(), '0A', 'correct first row rendered');
+      assert.notEqual(
+        table.getCell(table.rows.length - 1, 0).text.trim(), '99A', 'correct last row rendered'
```
billy-addepar: I think the message here does not reflect what it test. It should say "Last rendered row is not last data row"

tests/integration/components/basic-test.js
```diff
@@ -0,0 +1,103 @@
...
+      );
+
+      // scroll all the way down
+      await scrollTo('[data-test-body-container]', 0, 10000);
+
+      assert.notEqual(table.getCell(0, 0).text.trim(), '0A', 'correct first row rendered');
```
billy-addepar: same

tests/dummy/app/components/custom-footer.js
pzuraq: It's not needed for tests, but the template renders it and it wasn't there. I'd assumed it was meant to be and added it for future tests, because we'll likely want to verify that it works.

billy-addepar: This is a duplication of `footerValue` in `ember-table-footer`. I think I put it there in the template but did not actually use it. We can pass `footerValue=footerValue` in `ember-table-footer`.

pzuraq: I don't think we should do that just to reduce code. Going to merge this, we need to keep moving.

billy-addepar: That's good to me. We can clean up this later. I did not mean to bikeshed on this.

PR #487, opened 2018-02-12
Add accessibility test

billy-addepar:


billy-addepar:
@Addepar/ice

pzuraq:
I think this is a good basic start, based on what I'm reading about the addon, but I think we'll need to do much more thorough a11y testing in the future.

billy-addepar:
Agreed. We need more accessibility tests from this starting point.

PR #489, opened 2018-02-15
display: flex doesn't work well in IE11 and tables don'e render. Conditionally using display:block for IE10 and IE11 fixes this

jtabone16:
This is the CSS we're using in our app.

@media all and (-ms-high-contrast: none), (-ms-high-contrast: active) {
    // IE10+ CSS here
    .et-table {
      display: block;
    }
}


billy-addepar:
@jtabone16 thanks for pointing this out.

The table did support IE11 in the past. A recent commit broke the behavior on IE11. I have opened a new issue to track this:

https://github.com/Addepar/ember-table/issues/490

I will close this PR as the `cyril/ember-2.0` is not the correct branch for v2. master is the branch you should work on.

PR #492, opened 2018-02-27
Tests for sending action up after header reordering & resizing

billy-addepar:


tests/helpers/generate-table.js
```diff
@@ -88,7 +90,11 @@ export function generateColumns(columnCount, {
 const defaultActions = {
   onSelect(newRows) {
     this.set('selectedRows', newRows);
-  }
+  },
+
+  onColumnReordered() {},
+
+  onColumnReSized() {}
```
pzuraq: `onColumnResized`

addon/components/ember-table.js
```diff
@@ -154,6 +154,13 @@ export default class EmberTable extends Component {
   @type(optional(Action))
   onColumnReordered = null;
 
+  /**
+   * An action passed in the column reordering event
+   */
+  @argument
+  @type(optional(Action))
+  onColumnReSized = null;
```
pzuraq: `onColumnResized`

billy-addepar:
Method names are updated. The `onColumnResized` is actually used in one of the current action. I rename it to `onColumnReordering` to make it clearer

PR #499, opened 2018-03-30
LinkedListTree supports adding and removing rows dynamically

addepar-andy:
You can add or remove rows after the tree has been constructed.  The add/remove operations could be optimized, but we don't need that right now so i haven't done it.

This is on top of https://github.com/Addepar/ember-table/pull/498, so this will be a bit easier to review after the other merges.

Reviewers: @Addepar/ice

### DIFF
--- a/addon/components/tree-table.js
+++ b/addon/components/tree-table.js
@@ -24,7 +24,7 @@ export default class TreeTable extends EmberTable {
     if (row.collapse) {
       tree.expand(row);
     } else {
-      tree.collapseNode(row);
+      tree.collapse(row);
     }
   }
 }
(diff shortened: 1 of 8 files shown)


pzuraq:
@addepar-andy just a heads up, the plan is to make the LinkedListTree datastructure a private structure which wraps over a simpler tree object that gets passed to the table. We shouldn’t be manually doing tree manipulations like this in app code.

Will be back on Monday so I can help review then. Refactoring will probably take a while so definitely can get this merged to unblock you in the meantime, just wanted to let you know!

addepar-andy:
thanks chris!

pzuraq:
So the refactor work actually went much more quickly than expected, we need to move forward with the PA table replacements so we went ahead and merged it. I can pair with you next week to update your table to use it instead, it should be much easier, just find the node and do `pushObject`

PR #501, opened 2018-04-10
tree-table tree argument is nullable

addepar-andy:
reviewers: @Addepar/ice


billy-addepar:
If you do not have data for the tree, can you pass an empty array instead? I think an empty array is a better data than a null object.

addepar-andy:
i think `null` is a more consistent api: the `tree` property can either accept a single row (your root node) or `null`, if you don't have a root node.

billy-addepar:
Ah, that makes sense.

pzuraq:
I think it makes sense. It can also accept an array for the record, in the case where you have multiple "roots" (we don't use this but added it in because it's a fairly common use case, may be better for the open source world).

PR #504, opened 2018-04-12
 [refactor] Use position: sticky; for fixed headers/footers/columns

pzuraq:
This PR switches us to a rendering strategy using position: sticky;
in all modern browsers. This is much more performant and simpler in the
code, and maintains semantic table structure - everything we've been
after!

For IE11, the only browser we support which does not support sticky, we
use a rudimentary polyfill. The polyfill is not comprehensive, it only
solves our exact use case by using `translateX/Y` on scroll.

### DIFF
--- a/addon/components/ember-table.js
+++ b/addon/components/ember-table.js
@@ -3,7 +3,7 @@ import { assert } from '@ember/debug';
 
 import { action, computed } from '@ember-decorators/object';
 import { alias } from '@ember-decorators/object/computed';
-import { tagName, classNames } from '@ember-decorators/component';
+import { classNames } from '@ember-decorators/component';
 
 import { argument } from '@ember-decorators/argument';
 import { required } from '@ember-decorators/argument/validation';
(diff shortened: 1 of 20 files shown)


addon/utils/sticky-polyfill.js
```diff
@@ -0,0 +1,79 @@
...
+  component._footerResizeSensor = new ResizeSensor(
+    component._footer,
+    component._repositionHandler
+  );
+
+  positionStickyElements(component);
+}
+
+export function teardownStickyPolyfill(component) {
+  component.element.removeEventListener('scroll', component._repositionHandler);
```
billy-addepar: You might want to detach all resize sensors in the teardown.

addon/components/ember-table.js
```diff
@@ -246,18 +251,28 @@ export default class EmberTable extends Component {
...
     this.token.cancel();
 
+    if (IS_IE) {
+      teardownStickyPolyfill(this);
+    }
+
     super.willDestroyElement(...arguments);
   }
 
+  getChildElement(tagName) {
```
billy-addepar: This can be removed

billy-addepar:
🎉 

With this the table automatically supports text wrapping 😄

PR #507, opened 2018-04-13
[polyfill] Polyfill the behavior of position: sticky; in thead/tfoot

pzuraq:
The previous implementation of fixed headers/footers wouldn't actually
work for headers and footers that have more than 1 row because it fixed
td/th elements directly due to a bug in the Edge/Chrome versions of
sticky. This PR adds a polyfill which does several things:

* It positions the headers/footers correctly individually based
on a measurement of the row height
* It repositions whenever the row height changes by using a ResizeSensor
* It repositions whenever cells or rows are added by using
MutationObservers

This is definitely suboptimal, but it's much better than our previous
setup and should be temporary.

The polyfills have been moved into private, and the legacy polyfill has
been refactored to match the style of the new one. They're both class
based now and use a WeakMap to tie themselves directly to an element
rather than the component class itself. This means we don't dirty the
component (or modify its hidden class) and we and we keep all of our
implementation details completely private.

One final change was to add an explicit width on the table itself. The
previous hack of setting `width: 0;` doesn't actually work in IE11, it
collapses the entire table, so we need to actually set it to the real
width (as defined by the columns).


billy-addepar:
Do we have to reposition on every single cell? Is that possible to reposition for every row since it looks like all cells have the same offset?

addon/-private/sticky/table-sticky-polyfill.js
```diff
@@ -0,0 +1,105 @@
...
+
+    this.element.style.position = 'static';
+    this.side = element.tagName === 'THEAD' ? 'top' : 'bottom';
+
+    this.setupRaf = requestAnimationFrame(this.repositionStickyElements);
+
+    this.setupResizeSensors();
+    this.setupRowMutationObservers();
+
+    this.mutationObserver = new MutationObserver(() => {
```
billy-addepar: From what I understand, this observer reacts to every DOM event. In a complicated app, we would have lots of DOM events and this would be triggered multiple times. Why is using ResizeSensor not enough in this case?

pzuraq: It reacts to specifically to children being added and removed, that's what the `childList: true` option does here. We need to reposition if a node was added or removed, because that node has not yet been positioned and could affect the positioning of other nodes.

This allows us to have a generic polyfill without tying into Ember directly.

pzuraq:
Table rows have the same issue as `thead` and `tfoot`. According to the [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=702927) the underlying issue is that `position: relative;` is not supported on rows when it should be. Until this is fixed we must position on `th` and `td` directly.

PR #516, opened 2018-04-24
overflow scroll is auto

addepar-andy:
reviewers: @Addepar/ice

### DIFF
--- a/addon/styles/addon.scss
+++ b/addon/styles/addon.scss
@@ -1,6 +1,6 @@
 .ember-table {
   position: relative;
-  overflow: scroll;
+  overflow: auto;
   max-height: 100%;
   max-width: 100%;
   box-sizing: border-box;


pzuraq:
`overflow: scroll` is currently how we get fixed columns, headers, and footers. Turning this off will likely break things, as the sticky elements will instead default to whatever the parent scroll container is.  Why are you trying to change it?

addepar-andy:
because this creates scroll bars when the table when the table doesn't need them.  a table that doesn't need to scroll side to side (because you can already see all the columns) will still have the scrollbar, and you'll get ~10px of superfluous scrolling.

pzuraq:
That sounds like a bug with antiscroll and the way it inserts scrollbars in our app, native scroll bars shouldn't be visible if the content doesn't overflow as you're describing. 

I think it would be better to manually override the overflow style in your table for the time being and keep these styles here, otherwise the default behavior will be broken for users unless they know to make `.ember-table` a scroll container.

pzuraq:
I think this _should_ be fixed on master, if not let me know and we'll make fixing a priority.

Issue #505, opened 2018-04-12
API Refactor


pzuraq:
After some usage internally and experimentation, we've landed on a new API design that focuses on flexibility and sets us up for a better future API with contextual components.

Target API:

```hbs
{{#ember-table as |t|}}
  {{#t.head columns=columns sortings=sortings as |h|}}
    {{#h.row as |r|}}
      {{#r.cell as |column|}}
        {{column.name}}

        {{#if column.meta.isSorted}}
          <i class="sort-icon"></i>
        {{/if}}
      {{/r.cell}}
    {{/h.row}}
  {{/t.head}}

  {{#t.body rows=rows as |b|}}
    {{#b.row 
      class=(if b.rowContent.isSelected 'is-selected') 
      
      as |r|}}
      {{#r.cell 
        class=(if r.cellContent.isActive 'is-active') 
        
        as |value cell column row|}}
        {{value}}

        {{r.cellContent === cell}}
        {{r.rowContent === row}}
        {{r.columnContent === column}}
        {{r.value === value}}
      {{/r.cell}}
    {{/b.row}}
  {{/t.body}}

  {{#t.foot rows=rows as |f|}}
    {{#f.row as |r|}}
      {{#r.cell as |value cell column row|}}
        {{value}}
      {{/r.cell}}
    {{/f.row}}
  {{/t.foot}}
{{/ember-table}}
```

This design strikes a balance between brevity, expressiveness, and intuitiveness. Contextual components work as one would expect them to (unlike the previous cut with the `.component` suffix) and wrappers are discarded in favor of cleanly separated meta objects. There are two downsides:

1. Meta information is stored different betweens cells and rows/columns. `value` represents a cell's concrete value, which may be a primitive value (e.g. a string or number) and `cell` represents a cell's metadata. Contrast this with `row.meta` and `column.meta`, which exist as keys on the actual cell and row (which will be added by an object proxy). The main issue here is that `cell` is a made up construct - users pass us rows and columns and we use them to make "cells". Because the final value of a cell can be a primitive object, it's not possible to add the `meta` object in the same way.
2.  Unwrapped cells, rows, and columns are not available out of the final `td` or `cell` component. This will likely only be needed in a minority of cases, and an escape hatch is provided (being able to access the values on the API object at `${key}Content`.

Both of these issues can be taught around with proper documentation.

Desugared for <Ember 2.4:

```hbs
{{#ember-table as |t|}}
  {{#ember-thead api=t columns=columns as |h|}}
[...]

PR #517, opened 2018-04-26
[restructure] New Table API

pzuraq:
Refactors to the the broken out table API outlined in #505. Notes:

* Lots of APIs renamed and moved
  * `thead`
    * `columns`
    * `numFixedColumns` = `hasFixedColumn` (also now a boolean)
    * `onColumnResize` => `onResize`
    * `onColumnReordered` => `onReorder`
    * `columnMode` => `resizeMode`
    * `tableResizeMode` => `fillMode`
  * `tbody`
    * `tree`
    * `rows`
    * `estimateRowHeight`
    * `onSelect`
    * `selectedRows`
    * `selectionMode` => `selectMode`
  * `tfoot`
    * `tree`
    * `rows`
    * `estimateRowHeight`
    * `onSelect`
    * `selectedRows`
    * `selectionMode` => `selectMode`
* Checkboxes now render IFF `onSelect` exists AND `selectMode ===
  'multiple'`, which is a much easier heuristic to follow and enforces
  that the checkbox is always in the first cell
* Had to restructure the RowWrapper a bit to avoid backtracking rerender
* Headers/columns need to be reworked quite a bit, will be doing this in
  a followup PR


README.md
```diff
@@ -69,36 +69,38 @@ by all `valuePath` in `columns` array.
...
-  }}
-    {{#ember-table-row
-      row=row
-      as |cell|
-    }}
-      {{#component cell.column.cellComponent
-        cell=cell
-      }}
+  {{#ember-table as |t|}}
+    {{ember-thead api=t columns=columns}}
```
addepar-andy: maybe `table=t` instead of `api=t`?

pzuraq: Yeah I've been going back and forth on that one. On the one hand, it reads better for now, but on the other I think it may actually obfuscate what we're passing in and why. This especially becomes an issue later on:

```
{{#ember-tbody as |b|}}
  {{#ember-tr body=b as |r|}}
    {{#ember-td row=r as |value cell column row|}}
       ^r and row are not the same thing
    {{/ember-td}}
  {{/ember-tr}}
{{/ember-tbody}}
```

I think it may make sense to add the name as an alias to each component, so users can choose whichever one makes more sense to them and we can see which is less confusing over time. Once we get to 2.3 we'll be converting to contextual components anyways, so it's not really a big deal either way.

addon/components/-private/simple-checkbox.js
```diff
@@ -39,6 +38,11 @@ export default class SimpleCheckbox extends Component {
   @type(optional(Action))
   onChange = null;
 
+  @argument
+  @type('string')
+  @attribute('aria-label')
+  ariaLabel;
```
billy-addepar: What is this ariaLabel used for ?

addon/components/ember-tbody/template.hbs
```diff
@@ -0,0 +1,37 @@
...
+    as |proxy cells|
+  }}
+    {{#if hasBlock}}
+      {{yield (hash
+        selectRow=selectRow
+        rowValue=proxy
+        cells=cells)
+      }}
+    {{else}}
+      aoeu
```
billy-addepar: typo?

billy-addepar:
LGTM barring some comments

addon/components/-private/simple-checkbox.js
pzuraq: it's for the a11y test, checkboxes are now enabled based on `selectMode` so they're enabled for all tests. a11y test checks to make sure all form controls are properly labeled.

PR #521, opened 2018-05-14
[feat] ColumnTree

pzuraq:
This PR adds adds a ColumnTree data structure which manages the
hierarchy of columns and subcolumns. This allows us to have a nested
tree of columns with any configuration. All levels of columns can be
reordered and resized, including fixed columns on the left and the
right hand side.

Documentation has been added to the docs site for the new features, and
they are fully tested.


addon/-private/column-tree.js
```diff
@@ -0,0 +1,783 @@
...
+
+class TableColumnMeta extends EmberObject {
+  @readOnly('_node.isLeaf') isLeaf;
+  @readOnly('_node.isFixed') isFixed;
+  @readOnly('_node.width') width;
+  @readOnly('_node.offsetLeft') offsetLeft;
+  @readOnly('_node.offsetRight') offsetRight;
+}
+
+class TableColumnProxy extends ObjectProxy {
```
mixonic: What is the motivation for this proxy?

addon/-private/column-tree.js
```diff
@@ -0,0 +1,783 @@
...
+      set(meta, '_node', this);
+      meta.registerElement = this.registerElement.bind(this);
+
+      meta.startResize = tree.startResize.bind(tree, this);
+      meta.updateResize = tree.updateResize.bind(tree, this);
+      meta.endResize = tree.endResize.bind(tree, this);
+
+      meta.startReorder = tree.startReorder.bind(tree, this);
+      meta.updateReorder = tree.updateReorder.bind(tree, this);
+      meta.endReorder = tree.endReorder.bind(tree, this);
```
mixonic: Equivalent to `meta.endReorder = () => tree.endReorder(this)`. I prefer the closure but I don't have any convincing justification for changing the version already written here.

addon/-private/column-tree.js
```diff
@@ -0,0 +1,783 @@
...
+    this.token = new Token();
+
+    scheduler.schedule('sync', this.ensureWidthConstraint, this.token);
+
+    this.sortColumnsByFixed();
+    addObserver(this, 'columns.@each.isFixed', this.sortColumnsByFixed);
+    addObserver(this, 'widthConstraint', this.ensureWidthConstraint);
+  }
+
+  destroy() {
```
mixonic: On this `EmberObject` should you also be calling `this._super()` in `destroy`?

pzuraq: probably should, good catch!

addon/components/ember-tfoot/template.hbs
```diff
@@ -4,9 +4,14 @@
     rowIndex=rowIndex
     selectedRows=selectedRows
     columns=columns
+
+    onSelect=onSelect
+    selectMode=selectMode
     selectRow=selectRow
     toggleRowCollapse=toggleRowCollapse
 
+    metaCache=_rowMetaCache
```
mixonic: It seems like `_rowMetaCache` should not be underscored and "private" if it is passed to other components.

addon/components/ember-thead/component.js
```diff
@@ -163,315 +84,62 @@ export default class EmberTHead extends Component {
...
-    }
+    super.willDestroyElement(...arguments);
   }
 
-  @action
-  onColumnReorderEnded(columnIndex) {
-    if (this.get('hasSubcolumns')) {
-      // Disable column reordering when table has subcolumn.
+  fillupHandler = () => {
+    if (this.get('isDestroying')) {
```
mixonic: No need to use `get`/`set` with `isDestroying` or `isDestroyed` checks.

addon/components/ember-thead/template.hbs
```diff
@@ -1,33 +1,13 @@
...
+{{#each wrappedRows as |row index|}}
+  {{#if hasBlock}}
+    {{yield (hash
+      cells=row
+      _isHeader=true
+    )}}
+  {{else}}
+    {{ember-tr api=(hash
+      cells=row
+      _isHeader=true
```
mixonic: More semi-privateness to consider.

pzuraq: Yeah, my thinking was that ideally, this would never be exposed to users, and the underscore would give them the cue that this field should not be depended on, but in reality it is a public field.

Sidenote: Can't wait for true private fields taking away the ambiguity around what is "private" 😛

addon/-private/column-tree.js
mixonic: You may be able to model this as a WeakMap then, if it is a cache for each object in the system. I see that you're refactoring existing logic here, but proxies in general raise a performance and maintainability red flag for me.

pzuraq: Totally, I've gone back and forth on this over time as well. Taking a step back, I'd say there are three pretty decent APIs. For context, here is the current target API: 

```hbs
{{#ember-table as |t|}}
  {{#t.body as |b|}}
    {{#b.row as |r|}}
      {{#r.cell as |value cell column row|}}

      {{!-- value === the actual value of the cell, can be a primitive. row[column.valuePath] --}}
      {{!-- cell === the cell cache object, it's a proxy as well but a different kind designed to save memory and allocations, see CellProxy --}}
      {{!-- column === the column proxy, with added meta field --}}
      {{!-- row === the row proxy, with added meta field --}}

      {{/r.cell}}
    {{/b.row}}
  {{/t.body}}
{{/ember-table}}
```

What we're talking about is how we yield the inner API to users. I think this would be made much easier with named parameters, but in the meantime we do have to yield these positionally.

We could go a more verbose route:

```hbs
{{#r.cell as |cellValue cellMeta column columnMeta row rowMeta|}}

{{/r.cell}}
```

The common case is that users are _just_ using `cellValue` and `columnValue` so this may not be too bad. The other option we had was to _reverse_ the "proxy" behavior, so that the cell, row, and column were _wrapped_ by their proxies:

```
{{#r.cell as |cell column row|}}
  {{!-- cell === the cell cache --}}
  {{!-- cell.value === the value of the cell --}}

  {{!-- column === column meta --}}
  {{!-- column.value === the column itself --}}

  {{!-- row === row meta --}}
  {{!-- row.value === the row itself --}}

{{/r.cell}}
```

Alternatively, maybe we don't need to yield all this information at the cell level? If so, where do we yield it and why? Definitely open to suggestions, fresh eyes on this one would be helpful

PR #526, opened 2018-05-16
README: added instructions for documentation

addepar-andy:
found the local doc extremely helpful.  not sure if this is something we want to publicize yet?

reviewers: @Addepar/web-core


pzuraq:
This is good for now, once we get the actual gh pages site up we'll update the README.

PR #529, opened 2018-05-17
[bugfix] Plugs Memory leaks

### LINKED ISSUE #523: Memory leak
I find a consistent way to reproduce mem leak issue without switching route:
1) Goes to `scenarios/performance` route
2) Adds a checkbox in the template to hide/show table. This will destroy previous table and create a new one.
3) Scrolls down to middle of the table. Selects some row. Clicks on hide checkbox to hide it. Clicks it gain to show a new table. 
4) Repeat step 3 several times. 

Observe memory increase over time even with forced GC. I guess some of the event listener/observer is still being tied to the DOM even after table is destroyed.

pzuraq:
We had a number of memory leaks being caused primarily by nodes from the
ColumnTree and CollapseTree being referenced and un-garbage-collectable.
It turned out that this was due to their Ember Meta objects not being
destroyed - in certain cases, meta objects (in particular, chains) can
have dangling references.

This PR converts all objects to using EmberObject so we can clean up
their metas, and creates private caches for nodes in the trees so they
can clean up their children if they need to create a new set of nodes.

Also caught a memory leak caused by Hammer (gotta call `destroy`) and
a memory leak caused by the cell proxies in RowWrapper.

Fixes #523

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -76,17 +76,20 @@ class TableColumnMeta extends EmberObject {
 /**
   Single node of a ColumnTree
 */
-class ColumnTreeNode {
-  constructor(column, tree, parent) {
-    this.tree = tree;
-    this.column = column;
+class ColumnTreeNode extends EmberObject {
+  _subcolumnNodes = null;
+
+  constructor() {
+    super(...arguments);
+
+    let tree = get(this, 'tree');
+    let parent = get(this, 'parent');
+    let column = get(this, 'column');
 
     if (!parent) {
       this.isRoot = true;
     } else {
-      this.parent = parent;
-
-      let meta = getOrCreate(this.column, get(tree, 'columnMetaCache'), TableColumnMeta);
+      let meta = getOrCreate(column, get(tree, 'columnMetaCache'), TableColumnMeta);
 
       set(meta, '_node', this);
       meta.registerElement = (...args) => this.registerElement(...args);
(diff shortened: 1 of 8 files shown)


pzuraq:
You would think! But this is not an issue with our code, but Ember code itself as @mixonix and I discovered yesterday: https://github.com/emberjs/ember.js/blob/704fe834c9b7b2b42a721d9b4a63898b24f20644/packages/ember-meta/lib/meta.ts#L107

Chains in general can result in references which are not GC’d it seems. Yet another reason we should be relying less on chains and the old observation system and more on `@tracked`

billy-addepar:
🤦‍♂️

billy-addepar:
Wait, I do not see any data in table using this branch

screen shot 2018-05-17 at 10 36 00 am

addon/-private/column-tree.js
```diff
@@ -104,29 +107,47 @@ class ColumnTreeNode {
...
+      this._notifyCollection = () => notifyPropertyChange(parent, '[]');
+
       addObserver(this, 'maxChildDepth', this._notifyMaxChildDepth);
       addObserver(this, 'leaves.[]', this._notifyLeaves);
     }
   }
 
   destroy() {
-    removeObserver(this, 'maxChildDepth', this._notifyMaxChildDepth);
-    removeObserver(this, 'maxChildDepth', this._notifyLeaves);
```
ginomiglio, line 114: Did this introduce a different memory leak? I think the second line was wrong originally (probably meant it to be `removeObserver(this, 'leaves.[]', this._notifyLeaves);` but how are these observers getting cleaned up without them?

PR #530, opened 2018-05-18
column tree distributes width more safely

addepar-andy:
There is a bug where the width distributor will get into an infinite loop.  The loop would check for delta being zero, but if delta started as a decimal, then the check would never pass.  This does 3 things:

1. adds a loop guard so that the UI / tests don't freeze if something crazy happens to the loop.
2. rounds delta so the target is always an integer
3. makes delta chunks always integral and sum to the delta

Reviewers: @Addepar/web-core

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -19,6 +19,8 @@ const DEFAULT_COLUMN_WIDTH = 100;
 const DEFAULT_MIN_WIDTH = 50;
 const DEFAULT_MAX_WIDTH = Infinity;
 
+const LOOP_COUNT_GUARD = 500;
+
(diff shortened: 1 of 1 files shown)


addepar-andy:
i can write some unit tests for the new function i wrote, but i'm not really sure how to actually trigger this bug, so i'm not sure what kind of component test to write.

addepar-andy:
i think we should keep the flag in.  if something goes wrong on prod and guard is there, users will see an error.  without the guard, the entire tab will freeze.

pzuraq:
We can keep it in for now, but I’d like to circle back and add a few more assertions that give us guarantees before we release 1.0.0. It shouldn’t be possible to get in that state.

PR #532, opened 2018-05-19
[refactor] Move row meta to CollapseTree

pzuraq:
This PR refactors CollapseTree to take much more of the responsibility
of managing the state of rows, and setting up their meta objects. This
does a few things:

* Normalizes the tree structures - CollapseTree now behaves much like
  ColumnTree. At some point they could probably be made to extend from
  the same base class.
* Gives us enough context to correctly toggle selection (which allowed
  us to fix selection, and add the `grouping` mode)

### API Changes

* `ember-tbody` no longer accepts a `tree` object, it'll only take an
  array of rows. To get a single root node in a tree, pass in an array
  with a single parent node.
* `collapsed` -> `isCollapsed` on nodes to bring it inline with
  terminology through the repo
* New `selectMode` option: `grouping`
* Toggling collapse state is now done by a checkbox is the first cell.
  Both checkboxes and the table itself will be styled to match Addepar's
  internal table usage in an upcoming PR.
* Tree mode and collapsing are now toggleable via `enableTree` and
  `enableCollapse` on the `tbody`.

### Other Stuff

* Docs added for the `tbody`
* Lots of new tests added!

Note: There is a subtle memory leak introduced in this PR, would like to fix it before we merge but this one is really tricky. I've narrowed it down to the `onSelect` action, _somehow_ by sending rows up from there we're leaking.


addepar-andy:
> Toggling collapse state is now done by a checkbox is the first cell.

why a checkbox?  that isn't how the analysis table works

pzuraq:
We can style a checkbox to look just like the toggle icon that exists in our internal tables today. Using a checkbox is more semantically correct and gives us a lot of behavior and a11y for free.

PR #536, opened 2018-05-22
table row meta - dont select if destroying

addepar-andy:
came up in an integration test on iverson.  not really sure how to repro otherwise.

reviewers: @Addepar/web-core


addepar-andy:
confirmed that this fixes the iverson test

PR #537, opened 2018-05-23
[feat] Implement Column Sorting and Actions

pzuraq:
This PR implements column sorting with a default sort implementation,
and dropdown actions, which can be passed in at the header cell level.
Dropdown actions can be used to add buttons to the dropdown menu
associated with each column.

The sorting functionality is accomplished via merge sorting each level
of the CollapseTree. This also leaves the default sort intact, so users
can return to the default state.

Dropdown actions have been tested and documented, but the docs are not
exposed via navigation yet. Unsure of the API at this point.

Styles to come in the next PR.

### DIFF
--- a/addon-test-support/pages/-private/ember-table-header.js
+++ b/addon-test-support/pages/-private/ember-table-header.js
@@ -1,5 +1,9 @@
-import PageObject, { collection, hasClass } from 'ember-classy-page-object';
+import PageObject, { alias, collection, hasClass } from 'ember-classy-page-object';
 import { findElement } from 'ember-classy-page-object/extend';
+import { click } from 'ember-native-dom-helpers';
+
+import AddeDropdownPage from '@addepar/pop-menu/test-support/pages/adde-dropdown';
+import AddeSubDropdownPage from '@addepar/pop-menu/test-support/pages/adde-sub-dropdown';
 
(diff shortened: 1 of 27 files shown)


addepar-andy:
don't super understand the api here.  how would i turn off built in sorting?

pzuraq:
It depends on what you mean by "turn off". So far we're using a DDAU philosophy for modifying external data where possible (this isn't really possible with e.g. columns yet, because the data structure is so complicated). 

So, when you click on a table header cell, it sends the new sort order up in an action. It doesn't change any sort ordering itself yet, and if you don't set the new `sorts` in the parent context, it will not update. The external context retains control.

Taking it one step backward, if there _is no `onUpdateSorts` action, there is nothing to do when users click on the header. So, we only apply the sort actions and styles to header cell IF there is an `onUpdateSorts` action, and omitting it will "disable" sorting.

However, if you were to pass down `sorts` to the table, it would still actually sort! So you can disable user interaction, without disabling sorting itself. 

If that seems confusing, consider this: You've been asked to create a sort dropdown _external_ to the table, that changes the sort columns. This is supposed to replace the standard table interaction we included with the table, so you don't want to disable sorting, just provide a different (external) UX.

This API gives us that flexibility and makes the table component that much easier to use in different scenarios.

addepar-andy:
cool thanks!

PR #550, opened 2018-06-16
fix: example in why section

knownasilya:


pzuraq:
Failures should be fixed by #554, going to merge this

pzuraq:
Thanks!

PR #568, opened 2018-07-03
allow specifying the collection key

bgentry:
This helps the collection to better determine whether rows have been added, removed, or changed, and avoids unnecessary re-rendering when using non ember-data data libraries (or POJOs). You can see more info about it in the [vertical-collection docs](https://html-next.github.io/vertical-collection/#/settings):

> ```js
>   // This key is the property used by the collection
>   // to determine whether an array mutation is an
>   // append, prepend, or complete replacement. It is
>   // also the key that is passed to the actions, and
>   // can be used to restore scroll position with
>   // `idForFirstItem`.
>   //
>   // Note: `@identity` is a randomly generated value.
>   // If you want to save the id, use a unique property
>   // on your model (e.g. the `id` field on Ember Data
>   // models)
>   key: '@identity',
> ```

Since I use [ember-apollo-client](https://github.com/bgentry/ember-apollo-client) instead of ember-data, this is required in order to avoid re-rendering the entire collection every time a single item is changed.

Side note: there are probably other properties from vertical-collection that the user might want to be able to customize, so maybe they should be added too.

I couldn't figure out how this could be tested, but open to ideas if there's a way! Cheers :beers:

### DIFF
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -182,6 +182,17 @@ export default class EmberTBody extends Component {
   rows = [];
 
   /**
+    This key is the property used by the collection to determine whether an
+    array mutation is an append, prepend, or complete replacement. It is also
+    the key that is passed to the actions, and can be used to restore scroll
+    position with `idForFirstItem`. This is passed through to the
+    vertical-collection.
+  */
+  @argument({ defaultIfUndefined: true })
+  @type('string')
+  key = '@identity';
+
+  /**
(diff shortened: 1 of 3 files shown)


pzuraq:
No worries on testing this, it's just proxying behavior and part of composability best practices IMO is not testing functionality of other components. We have been slowly adding more proxy functionality for VC, each new option represents more API space though so we've been keeping it minimal when possible.

Thanks for this! Will be releasing the next beta soon, working on some other functionality at the moment as well.

PR #575, opened 2018-07-09
remove ember-faker

miguelcobain:
This PR removes ember-faker in favor of importing it from npm using ember-auto-import (since it also works in test code).
Also replaced a random number generation to use faker.


pzuraq:
Thanks for the PR!

PR #577, opened 2018-07-11
Check for isDestroyed before calling notifyPropertyChange

cah-danmonroe:
Fixes Assertion Failed: Cannot modify dependent keys for
`cellValue` on `<Ember.Object:ember###>` after it has been destroyed.

Demonstration of issue:
(https://ember-twiddle.com/aab786f62b838562163c11f8bee3a578?fileTreeShown=false&numColumns=0)


cah-danmonroe:
Verified #583 fixes the issue.

PR #579, opened 2018-07-13
use window scroll position for ember-router-scroll

bgentry:
When using the addon docs, I was getting annoyed that the documentation app did not scroll-to-top when clicking to a different docs section. This fixes that.

In the [ember-cli-addon-docs quickstart](https://ember-learn.github.io/ember-cli-addon-docs/latest/docs/quickstart#optional), they mention ember-router-scroll as an optional dependency to tackle this problem.

This installs the ember-router-scroll addon to properly remember scroll position between routes, and to reset it when the route changes. There was already a (copy-pasted?) config for this in the dummy app, but it pointed to an element that doesn't exist. Instead I just removed the config and it now uses the `window` scroll position instead, which works well.

### DIFF
--- a/tests/dummy/config/environment.js
+++ b/tests/dummy/config/environment.js
@@ -6,9 +6,6 @@ module.exports = function(environment) {
     rootURL: '/',
     locationType: 'history',
     historySupportMiddleware: true,
-    routerScroll: {
-      scrollElement: '#docs-viewer__scroll-body',
-    },
     fastboot: {
       hostWhitelist: [/^localhost:\d+$/],
     },


bgentry:
I'm noticing now that ember-cli-addon-docs actually [includes ember-router-scroll as a direct dependency](https://github.com/ember-learn/ember-cli-addon-docs/blob/b4bfe91ad58406808b960fc1870d571b16b462a4/package.json#L52) now. Perhaps its docs are out of date and we don't need to install it directly, only to make sure that it is configured? Gonna try that out to see if it was really just a problem due to bad config.

bgentry:
I updated the PR to merely fix the config for ember-router-scroll, which appears to work. Also opened a corresponding issue to fix the docs at https://github.com/ember-learn/ember-cli-addon-docs/issues/203.

pzuraq:
Thanks for the fix! Was wondering why that was happening

PR #583, opened 2018-07-18
[bugfix] Ensures computeds are properly destroyed

pzuraq:
The `willDestroy` hook doesn't appear to be an actual lifecycle event we can listen to anymore, so we need to hack around that by wrapping the original method.


cah-danmonroe:
I'll verify it fixes 577

cah-danmonroe:
Verified, this works for me.

pzuraq:
Awesome 😄 will release once your new feature is in as well!

PR #590, opened 2018-07-24
Adds prev, next,first, last computed properties to rowmeta

cah-danmonroe:
Will return the corresponding row relative to the current row.

Currently has an issue when rows are deleted and/or removed.  When rows are updated, the next time the new computeds are called, there is an assertion in the console:

`Assertion Failed: Cannot modify dependent keys for 'prev' on '<Ember.Object:ember###>' after it has been destroyed.`


pzuraq:
This looks great, thanks for taking it on! I'll see if we can reuse rowMeta values somehow so we don't run into the destroy issue

PR #596, opened 2018-07-26
failing test for selection after rows update

kturney:
This is a failing test for an issue we ran into around selecting items and then the table row content changing.

For our specific case, changing the sort triggers a network request, which replaces the rows in the table. If a user had selected some items and then changed the sort, they could no longer select more items.

However, the sort changing does not appear to be core to the issue. Just changing the rows is enough to reveal the bug.

I plan to dig in and see if I can find a fix, but any guidance would be appreciated 😄.

### DIFF
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -16,8 +16,16 @@ export const SELECT_MODE = {
   MULTIPLE: 'multiple',
 };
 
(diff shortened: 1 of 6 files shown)


pzuraq:
Thanks a ton! Failing tests really help pinpoint bugs quickly 😄 we'll get right on this

kturney:
Wow, that was fast! Thanks!

PR #605, opened 2018-08-07
Allow sorting on htmlSafe string

### LINKED ISSUE #569: Allow sorting on html safe string
Currently the sorting does not work when passing a value with HTML safe applied. 

`htmlSafe` is just a class with some methods on it. 

https://github.com/emberjs/ember.js/blob/v3.2.2/packages/ember-glimmer/lib/utils/string.ts#L61

Also, Ember has a method to check if a given value is html safe: `isHTMLSafe`.

josemarluedke:
Fixes #569.

### DIFF
--- a/addon/-private/utils/sort.js
+++ b/addon/-private/utils/sort.js
@@ -1,6 +1,10 @@
 import { compare, isNone } from '@ember/utils';
 import { get } from '@ember/object';
 
+function isHTMLSafe(str) {
+  return str !== null && typeof str === 'object' && typeof str.toHTML === 'function';
+}
+
 function merge(left, right, comparator) {
   let mergedArray = [];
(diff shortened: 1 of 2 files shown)


pzuraq:
It looks like the behavior of HTMLSafe is different for Ember 1, so we'll need to update this PR to account for that. You can use [ember-compatibility-helpers](https://github.com/pzuraq/ember-compatibility-helpers/) to have branching logic per version of Ember, should be included already

josemarluedke:
@pzuraq I have update the PR to include `isHTMLSafe` function. That would work with with Ember 1.x and newer versions.

josemarluedke:
Closing this as we have a been using a better approach by applying any any kind of formatting in the template instead of doing in the JS side.

PR #608, opened 2018-08-08
Yield to inverse when tbody rows are empty

ynotdraw:
One thing I noticed with this implementation is that since the yielded inverse content is rendered inside of the `tbody`, the content only spans the width of the first column by default. Any thoughts on the best way to solve this? Should it be wrapped in a row or something?

I'm not sure how you wanted to handle documentation either.  I could take a stab, if you want to point me in the right area where you think it makes sense (maybe under the `rows-and-trees` section? Does it deserve it's own section?).

Looks like I may need to do something for Ember 1 as well (SUPPORTS_INVERSE_BLOCK).

### DIFF
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -9,6 +9,8 @@ import { required } from '@ember-decorators/argument/validation';
 import { type, optional } from '@ember-decorators/argument/type';
 import { Action } from '@ember-decorators/argument/types';
 
+import { SUPPORTS_INVERSE_BLOCK } from 'ember-compatibility-helpers';
+
 import CollapseTree, { SELECT_MODE } from '../../-private/collapse-tree';
 
 import layout from './template';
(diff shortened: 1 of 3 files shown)


chrismllr:
+1 here! this would be great, to expose a "No results" state.

pzuraq:
Discussed this with @ynotdraw and I think it's a great feature, definitely want to get it merged!

It looks like there's a failure on the oldest Ember version we support. I can try to take a look, but I'm unsure when I'll have time. Once tests are passing there we can merge this.

ynotdraw:
I'm hoping to have some time this weekend to dig into the changes required to support the oldest Ember version with this PR 👍

ynotdraw:
@pzuraq howdy! I had some time to get back to these tests and support for Ember 1.x if you have some time to review!

twokul:
does it solve https://github.com/Addepar/ember-table/issues/631 or am I misreading the code?

ynotdraw:
@twokul oh nice - at the time that issue didn't exist, but yup, this PR should take care of that issue! 🎉

qpowell:
Thanks @twokul <3

jeannettetang:
<3

PR #619, opened 2018-09-09
Added possibility to set custom container selector

shydie:
Added possibility to set custom container selector.
In case if we don't need fixed height of `ember-table` and scrollable height need to be taken from some outer container, we must provide`{{vertical-collection}}` by container selector property. Also we provided table id as default selector, which is more unique than `.ember-table`

### DIFF
--- a/addon/components/ember-table/component.js
+++ b/addon/components/ember-table/component.js
@@ -96,6 +96,7 @@ export default class EmberTable extends Component {
     return {
       columns: null,
       registerColumnTree: this.registerColumnTree,
+      tableId: this.elementId,
     };
   }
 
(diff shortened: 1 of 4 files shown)


pzuraq:
I'm hesitant to merge this because I felt like this would be a larger issue to solve. We have the assumption baked into the code at the moment, that the table header and footer are contained within a scroll container. Just using a parent container for VC shouldn't be enough to fix all the other potential issues that'll pop up.

I'm also not able to work on this project at the moment, so if you need this functionality I'll merge it, but be aware that we were planning to add it and expected it to be more work.

shydie:
@pzuraq Yes, need this merge 'cause I'm blocked without these changes. btw I guess it's a necessary functionality to make table more flexible

Issue #609, opened 2018-08-14
RowMeta index becomes stale when a row is prepended


cyk:
_Thanks for all the great work on this -- abstracting flexible tables ain't easy._ 

Currently, the `rowMeta.index` property is only updated [whenever a node is accessed](https://github.com/Addepar/ember-table/blob/3756d030832de0e57d6bc5e25120a59255a47f76/addon/-private/collapse-tree.js). 

```js
  objectAt(index) {
    // ...

    // Set the perceived index on the meta. It should be safe to do this here, since
    // the row will always be retrieved via `objectAt` before being used.
    set(meta, 'index', index);

    return result;
  }
```

This seems sufficient for most cases. However, it doesn't cover scenarios where a row is prepended (or, actually, any non-append operation). It shifts the indexes for subsequent rows with no change to the row objects. So there is usually no need to retrieve the row.

My specific use case is rendering row numbers from `rowMeta.index`. The workaround is to force object retrieval by generating new objects, `this.get('rows').map(row => ({ ...row }));`. 

On Friday, I'm hoping to look into this a bit more and possibly come up with a solution. So any direction you can give would be much appreciated.

pzuraq:
The best way to solve this will be making `index` a computed property. In order to figure out what the index of a row, we'll need to give the row metas enough context. If we store the offset of each row relative to its parent, we should be able to walk back up the row meta tree and add the offsets together to get the final offset.

Does that make sense?

cyk:
Thanks for the direction. I'm working from [this failing test](https://github.com/cyk/ember-table/commit/0a11c01969d4af047ab9ffd0025fa6ea6f32f3b2#diff-4cd74aacae6cb66f5116c52d3c7ed739R176). My initial thought is to compute and set the row meta's relative offset as part of the `CollapseTreeNode.objectAt` routine.

To confirm, [this comment](https://github.com/Addepar/ember-table/blob/3756d030832de0e57d6bc5e25120a59255a47f76/addon/-private/collapse-tree.js#L702) refers to it as a "perceived index." This means the meta index should _not_ take into account collapsed nodes, right?

cyk:
After mulling it over a bit, here is what I've come up with. 

Some pseudo code for calculating offsets and index:

Prev offset:

- If there is an offsetList, use it (this handles any nodes)
- If there is no offsetList but there are children, use their length (this handles any leaf nodes)
- If there is no offsetList or children, use parentRow.children indexOf (this handles the root)

[...]

PR #623, opened 2018-09-14
[bugfix] Recompute row meta index when previous prepend causes shift

cyk:
Here is an attempt at solving for #609. When computing a row's meta index, the last known index (i.e., last objectAt()) is reconciled with its previous sibling's index. 

For example, if its previous sibling's index is equal to its own as a result of a prepend, then assume a shift is needed. This then cascades forward.


cah-danmonroe:
This PR also needs to update `prev` and `next` to use _lastKnownIndex
```
  @computed('_tree.length')
  get next() {
    let tree = get(this, '_tree');
    if (get(this, '_lastKnownIndex') + 1 >= get(tree, 'length')) {
      return null;
    }
    return tree.objectAt(get(this, '_lastKnownIndex') + 1);
  }

  @computed('_tree.length')
  get prev() {
    if (get(this, '_lastKnownIndex') === 0) {
      return null;
    }
    return get(this, '_tree').objectAt(get(this, '_lastKnownIndex') - 1);
  }
```

cyk:
@cah-danmonroe, I pushed a commit with those changes for `prev` and `next`. Thank you!

@twokul or @pzuraq when you get a moment, could you please review this? Is this a viable fix?

PR #626, opened 2018-10-03
[bugfix] Fix how selection works when selecting rows not yet inserted in the DOM 

gwak:
Because of how vertical-collection works, the TableRowMetaCache is only populated for a given row when inserted into the DOM, when vertical-collection invokes `objectAt(index)` function of tree node.
We thus need to setup non-existent meta cache from selected rows.


tempo22:
I can confirm this fix this issue for me aswell

efx:
@gwak the `package-lock.json` changes here should be removed. Quick tip is to just `git checkout master package-lock.json`.

efx:
I've verified this fixes a bug in our application as well. Thank you @gwak !

gwak:
> @gwak the `package-lock.json` changes here should be removed. Quick tip is to just `git checkout master package-lock.json`.

@efx I've just removed the `package-lock.json`

efx:
@gwak are you able to update this PR with master and see what the error is? Once that is done we can try to get this merged into master.

efx:
Additionally, a unit or integration test should be included to assert the change.

mixonic:
@gwak yeah I'm also very interested in learning how this manifests in usage.

bantic:
This should be fixed by #727, which has been released in `2.0.0-beta.9`.
@gwak 	When possible, can you try beta 9 and see if it fixes the issue for you?
I'm going to close this now as I believe #727 fully covers this, but please reopen if I'm mistaken.

PR #633, opened 2018-11-17
Replace `constructor` with `init`

twokul:
As per [RFC 337](https://github.com/emberjs/rfcs/pull/337), the way
classes are initialized has changed. Namely, arguments are no longer
available in `constructor` but they are still available in `init`.

Because we used `constructor` to set properties from passed in arguments
before, we started seeing error on Travis:

```
Error: Property set failed: object in path "unwrappedApi" could not be found.
```

This change switches to using `init` where appropriate and is a
backwards compatible change.

EDIT:

An unfortunate side-effect of replacing `constructor` with `init` is
that class properties are no longer assigned properly. They used to be
assigned after `init`. So all class properties initializers that are not decorated
with `@argument` need to be moved in `init`.

thanks to @pzuraq for brain dump and explaining me differences between `init` and `constructor`.

### DIFF
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -311,8 +311,8 @@ function closestLessThan(values, target) {
 class CollapseTreeNode extends EmberObject {
   _childNodes = null;
 
-  constructor() {
-    super(...arguments);
+  init() {
+    super.init(...arguments);
 
     let value = get(this, 'value');
     let parentValue = get(this, 'parent.value');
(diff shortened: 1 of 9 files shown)


josemarluedke:
@twokul Is it possible to get a new beta release containing these changes?

twokul:
@josemarluedke will do by the end of the week

twokul:
published via https://github.com/Addepar/ember-table/releases/tag/v2.0.0-beta.7

@josemarluedke

PR #637, opened 2018-12-04
Add optional containerWidthAdjustment to ember-thead component

kpfefferle:
# Problem

When using custom CSS-styled scrollbars on the `div.ember-table` element, the vertical scrollbar covers some of the container's width. In browsers such as Chrome, this causes `div.ember-table` to *also* display a horizontal scrollbar to view the `table` content that is now covered by the vertical scrollbar.

# Solution

In order to avoid this added horizontal scrollbar, @mixonic and I added here support for an optional argument to be passed to the `ember-thead` component called `containerWidthAdjustment`. This value gets passed along to the ColumnTree class to be applied to the calculated container width. This allows us to adjust the table width to account for the covering scrollbar in order to avoid a horizontal scrollbar in cases where we do not want one.

# Alternatives

As we explored this problem, we originally passed in a custom function for calculating the container width that itself performed the necessary adjustment. While using a custom function could possibly account for more complex use cases, we ultimately decided that it was overkill for our current (and the seemingly most common) need to adjust the table width by a set value.

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -654,7 +654,9 @@ export default class ColumnTree extends EmberObject {
       return;
     }
 
-    let containerWidth = getInnerClientRect(this.container).width * this.scale;
+    let containerWidthAdjustment = get(this, 'containerWidthAdjustment') || 0;
+    let containerWidth =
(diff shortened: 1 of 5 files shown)


kpfefferle:
@mixonic Good call. I've added an accompanying test and a note to the docs.

PR #646, opened 2018-12-25
[WIP] Initial upgrade to decorators state 2

josemarluedke:
This is an initial stab at upgrading to latest ember-decorators + babel 7.

This addon is breaking our app where another addon is using decorators state 2, probably because of the old dependency here of `@ember-decorators/babel-transforms`. 

Also, I need to confirm that the changes related to `@ember-decorators/argument` are correct. Essentially the option `defaultIfUndefined ` is not available anymore.


josemarluedke:
I'm getting the following error from `@ember-decorators/argument`.

```
object.js:20 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.getOwnPropertyDescriptor (<anonymous>)

```

screen shot 2019-01-04 at 5 46 26 pm

Is anyone more familiar with this to help out?

cyril-sf:
@josemarluedke Do you still get the same error if you rebase?

josemarluedke:
I think this can be closed as State 2 decorators are not moving forward. We can keep using stage 1 decorators. See https://github.com/emberjs/rfcs/pull/440.

josemarluedke:
It would still be nice to have babel 7 and etc. But it make sense to be a new PR.

Issue #638, opened 2018-12-10
I got error Cannot modify dependent keys for `cellValue` when I went to other route and back


bedrosenator:
I get data from the backend as ember data model and cast it to array for table. It works well on table's page. But when I try to go to some other page I see the error Assertion Failed: Cannot modify dependent keys for `cellValue` on `<(unknown):ember940>` after it has been destroyed

it happens here (in ember-table/-private/utils/computed)
``` javascript 
  _contentDidChange: Ember.observer('_content', function () {
      if (!this._isUpdating) {
        this._context.notifyPropertyChange(this._key);
      }
    }),
```
after this method here crash happens, because this.isDestroyed is true (ember-runtime/lib/mixins/observable)

``` javascript
  notifyPropertyChange: function (keyName) {
      (0, _emberMetal.notifyPropertyChange)(this, keyName);
      return this;
    },
```

Frozenfire92:
I'm also running into this issue. I can confirm that the code @bedrosenator highlighted is whats causing the problem (added logs before/after).

I tried blocking the call if `this.isDestroyed` or `this.isDestroying` is true to no avail

For some context:

My app has an `x-editor` component which contains a tabbed interface showing `source-table` (ember-table inside) component. 

```
{{#tabbed-interface tabs=tabs}}
  {{#if activeSource}}
    {{source-table source=activeSource}}
  {{/if}}
{{/tabbed-interface}}
```

Switching between tabs (activeSources) doesn't trigger the error
Leaving x-editor and coming back with a different context doesn't trigger the error
Leaving x-editor and coming back to the same context always triggers the error

note I'm not actually doing route transitions, leaving/entering x-editor happens from within a parent "display" component

ijlee2:
Hello. I don't suppose there has been progress to solving this issue? I ran into the problem today (using Ember Data and changing routes, as also seen by @bedrosenator) and am hoping for a fix soon.

Frozenfire92:
There is a fix in #648 but I suspect it will need #656 fixed first before it can merge

ijlee2:
@Frozenfire92 Cool, thank you!

PR #648, opened 2019-01-11
Don't try to notify the cell value of a change if the context is being destroyed

nolman:
I believe the fix is correct that the cellValue notifier shouldn't fire if the context is being destroyed. 

This fixes https://github.com/Addepar/ember-table/issues/638

### DIFF
--- a/addon/-private/utils/computed.js
+++ b/addon/-private/utils/computed.js
@@ -45,7 +45,7 @@ const ClassBasedComputedProperty = EmberObject.extend({
 
   // eslint-disable-next-line
   _contentDidChange: observer('_content', function() {
-    if (!this._isUpdating) {
+    if (!this._isUpdating && !this._context.isDestroyed && !this._context.isDestroying) {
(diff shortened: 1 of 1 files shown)


Frozenfire92:
Confirming that this fixes the issue follow up I reported in #638 

my issue was I was testing `this.isDestroyed/ing` instead of `this._context.isDestroying`

It appears the tests here are failing due to a yarn module being incompatible

Frozenfire92:
Bump, would be nice to have this move along in a release

jrcooke2:
Running into this same issue, would love to see this fix make it in soon!

cyril-sf:
@nolman would you mind rebasing against master? Hopefully this will be enough to get tests passing.

nolman:
@cyril-sf rebased and fixed a lint - I left a comment on #652 if you could take a look at that as well, would love to get off of our fork and back onto mainline :)

cyril-sf:
Thanks!

PR #657, opened 2019-02-11
Fixes #656 Stop caching yarn dependencies to prevent build inconsistencies

Frozenfire92:
This prevents Travis from caching to prevent build inconsistencies described in #656

Issue #656, opened 2019-02-08
Clear yarn cache on Travis (of incompatible sub-dependency)


cyk:
This build is [failing master](https://travis-ci.org/Addepar/ember-table/builds/487241773?utm_source=github_status&utm_medium=notification) and recent PRs (https://github.com/Addepar/ember-table/pull/652, https://github.com/Addepar/ember-table/pull/649, https://github.com/Addepar/ember-table/pull/648). This is caused by a sub-dependency specifying node compatibility >= 8:

```sh
error ip-regex@3.0.0: The engine "node" is incompatible with this module. Expected version ">=8". Got "6.16.0"
error Found incompatible module
```

The `ip-regex` is a sub-dependency of ember-cli-addon-docs. They ran into the same issue. The workaround was to [pin the related dependency](https://github.com/ember-learn/ember-cli-addon-docs/pull/302) to an earlier version that supports Node 6.

Though ember-table's version constraint for ember-cli-addon-docs includes the fixed version, it appears that Travis's yarn cache continues to have the incompatible version (i.e., `ip-regex@3.0.0`).

**TL;DR:** Clearing the yarn cache on Travis should resolve the issue.

Frozenfire92:
Is this best done via `.travis.yml` by disabling the cache?

https://github.com/Addepar/ember-table/blob/821270bdd1345e51c41de33e3f117199b0d11dca/.travis.yml#L14-L15

or via Travis interface for each build?

https://docs.travis-ci.com/user/caching/#clearing-caches

![](https://docs.travis-ci.com/images/caches-item.png)

Would love to get this going as #648 keeps popping up

cyk:
@Frozenfire92 Good question. I think  removing the yarn cache altogether is the better option. The reduced build time isn't worth (potentially) inconsistent builds.

bantic:
Recent builds at Travis seem to be passing without much issue, so I will go ahead and close this, but if I'm mistaken please feel free to reopen.

PR #658, opened 2019-02-21
Upgrade several project dependencies

josemarluedke:
cc/ @twokul


josemarluedke:
Tests are failing in `release` and `beta` channels. I believe these tests are actually broken because of changes in Ember v3.8 and above. I can confirm that `ember-table` does not work properly when used in an app with Ember v3.8.

cyk:
I think this upgrade work is running into the same blockers as the broken build fixing efforts. In short:

1. [Master](https://travis-ci.org/Addepar/ember-table/jobs/478606931#L516) then subsequent builds began to fail as a result of a sub-dependency (`ip-regex@3.0.0`) specifying incompatible Node engine requirements (#656)
2. No problem, the latest (within constraints) of ember-cli-addon-docs patched it (#657), Travis yarn cache is cleared to pull it in
3. This uncovered a conflict with ember-cli-addon-docs ember-data dependency which manifests as [`Global error: Uncaught TypeError: Cannot read property 'prototype' of undefined`](https://travis-ci.org/Addepar/ember-table/jobs/491781352#L587) (also [last failure for this PR](https://travis-ci.org/Addepar/ember-table/jobs/496274530#L601))
4. No worries, ember-cli-addon-docs v0.6.0 patches it, so a [commit was added to #657](https://github.com/Addepar/ember-table/pull/657#issuecomment-463803992) that bumps the version. 

Finally, the [latest failure on #657](https://github.com/Addepar/ember-table/pull/657#issuecomment-465126681) is the same this branch experiences when upgrading ember-cli-addon-docs (sass import failures of tailwind styles). Thanks for reporting it in https://github.com/ember-learn/ember-cli-addon-docs/issues/326.

FWIW: I suspect that [v0.6.0](https://github.com/ember-learn/ember-cli-addon-docs/releases/tag/v0.6.0) introduced some breaking change in addition to the fix, as pinning directly to https://github.com/ember-learn/ember-cli-addon-docs/commit/97dcf3bc8478e45c217a38d15e2117df2e18c32e seems unaffected.

josemarluedke:
@cyk What is the path forward on this then? This PR as it is, is passing (with disabled ember beta + release). 

This is just one step closer to get other dependencies up to date. We are currently blocked by `ember-table` because of its old dependency of ember-decorators, which is my primary driving motivation for this PR.

I hope we can get some progress on updating `ember-table` to latest dependencies, mainly `ember-decorators`. 🤞

cyk:
@josemarluedke, I'd like to clarify that my previous comment means to corral related efforts and increase visibility. I think your work here is very helpful. I'm looking forward to the ember-decorator upgrades, so thank you. 🙂 

As far as the path forward, it's really a maintainers call. The two issues here are **1)** builds failing from a sub-dependency's breaking change (#656) and more recently, **2)** builds failing from incompatibilities with latest Ember (#659). 

**My 2¢** — It might be helpful to tackle the first issue with a minimal set of changes before diving into the second. We understand the breakages for the former; the latter requires more investigation and is possibly more involved. 

This makes the cause of what broke it in the first place clear and gives us a reliable base to either proceed with remaining dependency upgrades and/or fix Ember incompatibilities. Also, unblocks other PRs that might not need >= 3.8 compatibility.

If it's acceptable to allow failures for incompatible versions of Ember, then we might consider #657 fix. (We'd have to update it with reverting the last commit and adding allowed CI failures.) Of course, the yarn cache clearing is still an option as well.

cyril-sf:
@josemarluedke Thanks for the help.

Can you rebase against master to resolve the conflict?

josemarluedke:
@cyril-sf Just pushed a rebase. Let's see if tests pass.

cyril-sf:
Thanks for the help!

PR #661, opened 2019-03-08
Fix build

cyril-sf:
Travis has been reporting that the library doesn't pass tests for a while.


cyril-sf:
Weird errors https://travis-ci.org/Addepar/ember-table/builds/504176533

PR #662, opened 2019-03-10
[FEATURE] Add support for context menu on header cells

cyril-sf:
``` {{ember-th api=r onContextMenu='actionHandler'}} ```

### CODE REVIEW

File: addon/components/ember-th/component.js [hunk 1]
```diff
@@ -162,6 +170,13 @@ export default class EmberTh extends BaseTableCell {
     }
   }
 
+  contextMenu(event) {
+    this.sendAction('onContextMenu', {
```
twokul: why are we sending an object w/ the action as oppose to just sending an `event`?

cyril-sf: Copy/paste and a change of mind while testing (I debated sending part of the API/meta) 😄 

Good catch

cyril-sf: @twokul fixed, ready for review

PR #664, opened 2019-03-13
[CHORE] Update ember-decorators

cyril-sf:


cyril-sf:
So this fails with Ember 1.12 because of ember-decorators/ember-decorators#148

Screen Shot 2019-03-26 at 5 25 37 PM

When `uniqBy` was added, some code was present to make sure it would not break with older versions of Ember. This protection was removed with ember-decorators/ember-decorators#308.

We can't use `ember-decorators` >= 3.1.0 without patching it.

mixonic:
@cyril-sf what is the motivation for upgrading to `ember-decorators >= 3.1.0`, for context?

cyril-sf:
At the very least, to keep our dependencies up to date.

ezpuzz:
updating ember-decorators to the latest would unblock me from upgrading ember-cli-typescript in my app because having multiple versions in the dependency tree currently creates a runtime conflict and manually resolving to the newer or older is incompatible.   newer ember-cli-typescript is much faster on my project.

cyril-sf:
As we dropped support for `ember-decorators`, I'm closing this PR.

PR #665, opened 2019-03-13
Add yarn.lock

cyril-sf:


.travis.yml
```diff
@@ -41,7 +41,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
```
kpfefferle: Yarn lockfiles are not respected by consuming apps, so if we want to use a lockfile for development (which I'm in favor of), we should still let CI ignore the lockfile as this will more closely reflect the sub-dependencies that a consuming app would get when it installs the addon.

See the recommended addon blueprint here: https://github.com/ember-cli/ember-cli/blob/6f2d49a341399d255b40e2bd832b31b7a9bdf8c3/blueprints/addon/files/.travis.yml#L68

cyril-sf: Good point

.npmrc
```diff
@@ -0,0 +1 @@
+@addepar:registry = "https://registry.npmjs.org/"
```
twokul: do we need these files?

cyril-sf: Yes, otherwise it will use our private registry in the yarn.lock file for all the `@addepar/` packages

PR #668, opened 2019-03-17
[FIX] yarn.lock is outdated and adds test coverage

jelhan:


mixonic:
@jelhan thank you for taking a look at this! I'm going to close this PR for now since we have mutated the `yarn.lock` a few times recently. I hope it is now in sync. Adding `yarn check` sounds good though!

PR #669, opened 2019-03-20
ember-th renders only the block when one is passed.

cyril-sf:
It will allow to totally customize how it looks.

This will require templates to be updated if they were relying on the old behavior.
I'd also like to provide an easy way to reuse the existing html/logic to make that transition easier.

### DIFF
--- a/addon/components/ember-th/sort-indicator/component.js
+++ b/addon/components/ember-th/sort-indicator/component.js
@@ -1,0 +1,57 @@
+import Component from '@ember/component';
+import layout from './template';
+
+import { argument } from '@ember-decorators/argument';
+import { required } from '@ember-decorators/argument/validation';
+import { type } from '@ember-decorators/argument/type';
+import { tagName } from '@ember-decorators/component';
+import { readOnly } from '@ember-decorators/object/computed';
+
+/**
+  The table header cell sort indicator component. This component renders the state of the sort on the column (ascending/descending/none).
+
+  ```hbs
+  <EmberTable as |t|>
+    <t.head @columns={{columns}} as |h|>
+      <h.row as |r|>
+        <r.cell as |columnValue columnMeta|>
+          {{columnValue.name}}
+
+          <EmberTh::SortIndicator @columnMeta={{columnMeta}} />
+        </r.cell>
+      </h.row>
+    </t.head>
+
+    <t.body @rows={{rows}} />
+  </EmberTable>
+  ```
+  @yield {object} columnMeta - The meta object associated with this column
+*/
+
+@tagName('')
+export default class EmberThSortIndicator extends Component {
+  layout = layout;
+
+  /**
+    The API object passed in by the table header cell
+  */
+  @argument
+  @required
+  @type('object')
+  columnMeta;
+
+  @readOnly('columnMeta.isSortable')
+  isSortable;
+
+  @readOnly('columnMeta.isSorted')
+  isSorted;
+
+  @readOnly('columnMeta.isSortedAsc')
+  isSortedAsc;
+
+  @readOnly('columnMeta.isMultiSorted')
+  isMultiSorted;
+
+  @readOnly('columnMeta.sortIndex')
+  sortIndex;
+}
--- a/addon/components/ember-th/template.hbs
+++ b/addon/components/ember-th/template.hbs
@@ -2,21 +2,8 @@
   {{yield columnValue columnMeta}}
 {{else}}
   {{columnValue.name}}
-{{/if}}
 
-{{#if isSorted}}
-  <span data-test-sort-indicator class="et-sort-indicator {{if isSortedAsc 'is-ascending' 'is-descending'}}">
-    {{#if isMultiSorted}}
-      {{sortIndex}}
-    {{/if}}
-  </span>
-{{/if}}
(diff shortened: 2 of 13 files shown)


mixonic:
@cyril-sf an option here might be to have the sorting logic put into a component, and have the `ember-th` yield that component for optional invocation by the block. That would provide a fairly easy adoption path for this change.

As there is no stable release I think this is ok as a breaking change. We need to fix 3.x compat before making a stable release IMO, but that is the only thing that is a blocker.

kpfefferle:
While I like this idea in theory, I'm worried about this change breaking tables in our app that have already been converted to ET2 and expect the previous block behavior. When we feel like this has settled into a theoretical ideal, we need to compare it to the tables that have already been converted to ET2 in our own app and look for a good migration strategy.

addon/components/ember-th/template.hbs
```diff
@@ -2,21 +2,8 @@
   {{yield columnValue columnMeta}}
 {{else}}
   {{columnValue.name}}
-{{/if}}
-
-{{#if isSorted}}
```
mixonic: The corresponding properties on the ember-th component instance should probably be dropped?

For example: https://github.com/Addepar/ember-table/blob/master/addon/components/ember-th/component.js#L78

cyril-sf: done

addon/components/ember-th/sort-indicator/component.js
```diff
@@ -0,0 +1,57 @@
...
+  The table header cell sort indicator component. This component renders the state of the sort on the column (ascending/descending/none).
+
+  ```hbs
+  <EmberTable as |t|>
+    <t.head @columns={{columns}} as |h|>
+      <h.row as |r|>
+        <r.cell as |columnValue columnMeta|>
+          {{columnValue.name}}
+
+          <EmberThSortIndicator columnMeta=columnMeta>
```
mixonic: I think this should be `<EmberTh::SortIndicator @columnMeta={{columnMeta}} />`

tests/dummy/app/pods/docs/guides/main/table-customization/template.md
```diff
@@ -42,6 +42,45 @@ lookup paths:
...
+          @widthConstraint='gte-container'
+          @fillMode='first-column'
+        
+          as |h|
+        >
+          <h.row as |r|>
+            <r.cell as |columnValue columnMeta|>
+              {{ember-th/sort-indicator columnMeta=columnMeta}}
+              Custom Header {{columnValue.name}}
+              {{ember-th/resize-handle columnMeta=columnMeta}}
```
mixonic: These two `{{` invocations for the sort/resize components should probably also be using angle invocation.

```hbs
<EmberTh::SortIndicator @columnMeta={{columnMeta}} />
Custom Header {{columnValue.name}}
<EmberTh::ResizeHandle @columnMeta={{columnMeta}} />
```

tests/dummy/app/pods/docs/guides/main/table-customization/template.md
```diff
@@ -42,6 +42,45 @@ lookup paths:
   {{demo.snippet name='table-customization-custom-cell-template.js' label='component.js'}}
 {{/docs-demo}}
 
+If you want to customize a header cell but still want to include the elements to sort and
+to resize a column, you can use the `ember-th/sort-indicator` and `ember-th/resize-handle`
```
mixonic:
```suggestion
to resize a column, use the `ember-th/sort-indicator` and `ember-th/resize-handle`
```

addon/components/ember-th/sort-indicator/component.js
cyril-sf: 👍 I never used that form and couldn't find anywhere in the docs how it's supposed to be.

tests/dummy/app/pods/docs/guides/main/table-customization/template.md
cyril-sf: 👍

tests/integration/components/headers/ember-th-test.js
```diff
@@ -0,0 +1,60 @@
...
+        {{/ember-th}}
+      {{/ember-tr}}
+    {{/ember-thead}}
+
+    {{ember-tbody api=t rows=rows}}
+  {{/ember-table}}
+  `);
+  await firstHeader.click();
+
+  assert.equal(this.$('th [data-test-block]').length, 2, 'Header cells render passed block');
```
rondale-sc: I think mixing the page object selectors with regular selectors feels a bit weird.  Is there a way we could use the page object everywhere or not at all?

cyril-sf:
> @cyril-sf Is this missing tests for sort indicator and resize handle?

As it's mostly a code refactor, and no other unit tests exist I didn't write any. Happy to add some.

tests/integration/components/headers/ember-th-test.js
cyril-sf: Good catch 👍

mixonic:
@cyril-sf sort indicator and resize handle are new public API, as such I expect they would be tested so we don't break them. I agree while they were implementation details they were better tested via their integration, but at least a basic set of "this public API exists" tests seem appropriate.

PR #671, opened 2019-03-29
[CHORE] Node 6 support - pin jsdom

cyril-sf:
Screen Shot 2019-03-28 at 4 53 33 PM

```error jsdom@14.0.0: The engine "node" is incompatible with this module. Expected version ">=8". Got "6.17.0"```


mixonic:
🙈 Oh man I wonder what jsdom is even doing as a dependency 😬

PR #674, opened 2019-04-01
[CHORE] Pin ember-cli-page-object

cyril-sf:
Fix build issue on Travis
https://travis-ci.org/Addepar/ember-table/jobs/513359138

Related to san650/ember-cli-page-object#446

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
   },
   "homepage": "https://Addepar.github.io/ember-table",
   "resolutions": {
+    "ember-cli-page-object": "1.15.1",
     "jsdom": "^11.11.0"
   }
 }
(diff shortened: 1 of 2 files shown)


mixonic:
It seems weird that the resolution did not change the ember-cli-page-object version in the yarn.lock, although there are a lot of unrelated changes.

cyril-sf:
@mixonic without it, Travis (or anything not relying on the `yarn.lock` file) would grab the latest version of `ember-cli-page-object` (v1.15.2 https://github.com/san650/ember-cli-page-object/releases) which breaks Ember Table tests.

cyril-sf:
For information, the Page Object for Ember Table is built to be used by applications using Ember Table.

mixonic:
@cyril-sf seems fine, though [ember-classy-page-object](https://github.com/Addepar/ember-table/blob/master/package.json#L64) is a devDependency. If consumers are expected to use it I would think it would be a dependency 🤔 Unless I am misunderstanding how that would work.

Maybe this is not an issue for Addepar downstream since we have a direct dependency on `ember-classy-page-object`.

PR #675, opened 2019-04-01
[BUGFIX] Empty values are sorted properly

cyril-sf:


mixonic:
Good catch 👍 and test.

PR #676, opened 2019-04-04
widthConstraint is set on the table

cyril-sf:
The `widthConstraint` option decides how the table fits its container.
It makes sense to set it on the table and not on the header.
`widthConstraint` is now available to the `ember-thead` component throught the api object passed.

This also allows to set a `et-eq-container` class on the table.

```hbs
<EmberTable @widthConstraint='eq-container' />
```

results in

```html
<table class='et-eq-container'>
</table>
```


cyril-sf:
Compatibility issue with `EmberTHead` definition to be looked at (all demos fail).

mixonic:
Dismissing this as stale

PR #677, opened 2019-04-08
Start converting to classic classes

rwwagner90:
I wanted to try to start converting back to classic classes, but the tests are not happy. Do we need to convert everything to classic before tests will run?

### DIFF
--- a/addon/-private/utils/default-to.js
+++ b/addon/-private/utils/default-to.js
@@ -1,0 +1,37 @@
+import { computed } from '@ember/object';
+
+let VALUES = new WeakMap();
+
+function valuesFor(obj) {
+  if (!VALUES.has(obj)) {
+    VALUES.set(obj, Object.create(null));
+  }
+
+  return VALUES.get(obj);
+}
+
+export default function defaultTo(defaultValue) {
+  return computed({
+    get(key) {
+      let values = valuesFor(this);
+
+      if (!(key in values)) {
+        values[key] = typeof defaultValue === 'function' ? defaultValue() : defaultValue;
+      }
+
+      return defaultValue;
+    },
+
+    set(key, value) {
+      let values = valuesFor(this);
+
+      if (value === undefined) {
+        values[key] = typeof defaultValue === 'function' ? defaultValue() : defaultValue;
+      } else {
+        values[key] = value;
+      }
+
+      return values[key];
+    },
+  });
+}
--- a/addon/components/ember-table/component.js
+++ b/addon/components/ember-table/component.js
@@ -78,23 +78,22 @@ export default Component.extend({
       }
     }
 
-    super.willDestroyElement(...arguments);
+    this._super(...arguments);
   },
 
   tableStyle: computed('tableWidth', function() {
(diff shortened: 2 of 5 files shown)


rwwagner90:
@mixonic nothing really, other than since Addepar needs 1.x and things are not working with latest decorators, it's better to go ahead and convert back to classic classes. I am not sure why this is making a bunch of tests fail, but I hope to sync with @pzuraq about it tomorrow.

mixonic:
@rwwagner90 🤔 I'll try to hang in #ember-table on Discord this week. FWIW Addepar is actively using this project from `master` and we're not blocked by ES classes being used as they currently are in this repo.

IMO the only blocker on releasing a 1.0 for this library is having support for the latest version of Ember (without breaking backwards compatibility of course). Everything won't be perfect, but a Addepar is shipping it in production it seems like a good time to start taking API stability seriously. We can rally contributors, bugfixes and improvements on top of the 1.0 release.

rwwagner90:
@mixonic it is my understanding, after talking with @pzuraq that the latest decorators stuff is not compatible with earlier Ember versions, and without the latest decorators, any apps on 3.10+ will have errors trying to use ember-table. @pzuraq can you confirm that?

pzuraq:
So, to clarify the exact issues:

1. Ember 3.10.0 will be updating its CPs to a very different internal implementation than what exists currently to support decorators directly
2. `ember-decorators` v5 released a bugfix version to maintain compatibility, but we've also added deprecations to the APIs that will be replaced by master, and we don't have a way of targeting those deprecations to just addon code and not app/user code.
3. `ember-decorators` stopped testing against Ember v1 some time ago, I'm unsure which version that happened in though (looks like we didn't log it in the release notes)
3. `ember-decorators-polyfill` does not have the same support matrix as `ember-decorators`, and while it could be possible to add support, it would be for versions of Ember where we explicitly do _not_ recommend native class syntax or decorator usage anymore. They _should_ work similarly to v2, and _may_ work on Ember 1.12, but we aren't testing that far back anymore.

With that in mind, and the support matrix this addon requires, I thought it made most sense to convert this addon to classic classes, and move forward to native classes again once you are able to reach a version that is fully supported by the native class polyfill (ideally we'll get support for 2.18 soon). Since I was the one that built this addon out with native classes, I'm happy to help with the conversion! It really was a great way to figure out the nuances of native classes and helped us a lot in the design process.

If y'all still do want to push forward with native classes, I'm unsure how much support we'll be able to give you. We can absolutely try to make it all work, but we're also spread pretty thin trying to move forward with Octane 😕

pzuraq:
Also, FWIW, the conversion from classic classes _back_ to native classes should be completely codemoddable by the time we're done with it and you're ready to adopt them, so it shouldn't be nearly as painful to go back later on 😄

pzuraq:
Thanks for getting this started @rwwagner90 !

cyril-sf:
@pzuraq @rwwagner90 what's the consequence for the autogenerated API doc?

pzuraq:
The comment tags should still pass information through I believe, if not it should be possible to update the ESDoc adapter to pass the information through.

PR #679, opened 2019-04-11
[BUGFIX] Ensure dynamicAlias works on latest Ember

pzuraq:
There were some subtle changes on the latest Ember that made aliases
and properties that include . in their name to stop working. Instead
of creating dynamic aliases, we now just access the properties directly
on _context, which should fix the issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - "8"
+  - '8'
 
 sudo: required
 dist: trusty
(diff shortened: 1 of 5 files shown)


mixonic:
@pzuraq I'm keeping an eye on https://travis-ci.org/Addepar/ember-table/builds/518906233?utm_source=github_status&utm_medium=notification where `release` etc are building. Should this also add a recent version of Ember to the build?

rwwagner90:
@mixonic shouldn't ember-release, etc be the recent versions of Ember?

mixonic:
@rwwagner90 Yes. I meant to suggest that an Ember version be added to the standard versions tested instead of as an "allowed failure".

Regardless there remains a failure, though it almost looks like something in the test suite from a quick look.

josemarluedke:
@mixonic Any change we can get a new beta containing this change?

PR #680, opened 2019-04-13
thead, th, and tr to classic components

nummi:


rwwagner90:
@pzuraq I think I fixed all but one test here. If you happen to know what might be the issue with that one let me know, but otherwise @nummi is going to look into that one this week.

addon/components/ember-thead/component.js
```diff
@@ -36,77 +31,81 @@ import layout from './template';
...
 
   /**
     An optional sort
+    @argument({ defaultIfUndefined: true })
+    @type(optional(Function))
   */
-  @argument({ defaultIfUndefined: true })
-  @type(optional(Function))
-  sortFunction = sortMultiple;
+  sortFunction: sortMultiple,
```
pzuraq: Should also likely be `defaultTo`

addon/components/ember-th/component.js
```diff
@@ -39,95 +32,84 @@ const COLUMN_REORDERING = 2;
...
   */
-  @className
-  @readOnly('columnMeta.isReorderable')
-  isReorderable;
+  isReorderable: readOnly('columnMeta.isReorderable'),
 
-  @attribute('colspan')
-  @readOnly('columnMeta.columnSpan')
-  columnSpan;
+  columnSpan: readOnly('columnMeta.columnSpan'),
```
pzuraq: This needs to be added to `attributeBindings`

addon/components/ember-th/component.js
```diff
@@ -39,95 +32,84 @@ const COLUMN_REORDERING = 2;
...
 
-  @attribute('rowspan')
-  @readOnly('columnMeta.rowSpan')
-  rowSpan;
+  rowSpan: readOnly('columnMeta.rowSpan'),
```
pzuraq: This needs to be added to `attributeBindings`

addon/components/ember-thead/component.js
```diff
@@ -36,77 +31,81 @@ import layout from './template';
...
 
   /**
     An ordered array of the sorts applied to the table
+    @argument({ defaultIfUndefined: true })
+    @type(optional(Function))
   */
-  @argument({ defaultIfUndefined: true })
-  @type(optional(Function))
-  compareFunction = compareValues;
+  compareFunction: compareValues,
```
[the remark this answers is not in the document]
rwwagner90: I tried adding `defaultTo` and had lots of failures.

addon/components/ember-thead/component.js
rwwagner90: Same here

addon/components/ember-th/component.js
rwwagner90: Good catch!

addon/components/ember-th/component.js
rwwagner90: 👍

rwwagner90:
@pzuraq I added the missing `attributeBindings` and all tests pass now 🎉 

I tried the `defaultTo` on those fields you mentioned, and it blew up. Got a lot of things like `sorts is not iterable` IIRC.

addon/components/ember-thead/component.js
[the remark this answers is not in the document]
pzuraq: ah, it's trying to run it because it's a function, it treats it like an initializer. So, kind of counterintuitive, but you'd have to do:

```js
defaultTo(() => compareValues),
```

If no tests are failing maybe it's not an issue, but it definitely would be different than the way other args are defaulted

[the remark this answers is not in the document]
rwwagner90: I tried that and it still throws the errors. I'm curious, why do we need `defaultTo`? Wouldn't what we set be the default, and then anything passed to the component override it?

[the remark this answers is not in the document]
pzuraq: It covers two cases:

1. When you pass `undefined` as the argument. This typically happens as a result of _prop drilling_ when you're trying to wrap a component and are forwarding all of the arguments from the parent. The parent doesn't have defaults defined, so it defaults to `undefined`, which then overwrites your default.

2. When you _update_ the argument to be `undefined` for some reason. This is much less common, but does happen. If you update the argument in the external space to be undefined, maybe because you weren't thinking about what it would do to your children, that's typically not the behavior you want. The CP will provide the default instead.

We use the prop drilling in tests pretty frequently, because we have a default table template that has every prop value passed into it. I'll see if I can get it to work.

rwwagner90:
@pzuraq to reiterate what I said in discord here:

I think I figured out a way to make it work, but not sure if it is correct

```js
if (sorts) {
    for (let { valuePath, isAscending } of sorts) {
```

Had to check that sorts was a thing before using `of sorts`.

I guess the computed from `defaultTo` gets executed more than once, and it is a timing thing.

Please let me know if this sounds like the correct solution or not.

PR #682, opened 2019-04-16
Convert collapse-tree, column-tree, and ember-td to classic

rwwagner90:
cc @pzuraq


mixonic:
Thanks for pushing this forward @rwwagner90.

rwwagner90:
You're welcome @mixonic 😃

PR #684, opened 2019-04-23
[BUG] Setting the column width to its current value works

cyril-sf:
It used to return `undefined` which would lead the `delta` to be `NaN`,
causing an infinite loop when resizing a leaf column.

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -359,19 +359,21 @@ const ColumnTreeNode = EmberObject.extend({
       delta = Math.max(Math.min(oldWidth + delta, maxWidth), minWidth) - oldWidth;
 
       if (delta === 0) {
-        return;
+        return oldWidth;
       }
 
       if (get(this, 'isLeaf')) {
         let column = get(this, 'column');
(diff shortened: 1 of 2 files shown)


addon/-private/column-tree.js
```diff
@@ -359,7 +359,7 @@ const ColumnTreeNode = EmberObject.extend({
       delta = Math.max(Math.min(oldWidth + delta, maxWidth), minWidth) - oldWidth;
 
       if (delta === 0) {
-        return;
+        return oldWidth;
```
mixonic: Further down in two places:

```js
return set(column, 'width', oldWidth + delta);
```

Should also be:

```js
let width = oldWidth + delta;
set(column, 'width', width);
return width;
```

Relying on the return value from `set` is a bit weird (and behaves differently in some version of Ember 3.x iirc). Better to just return the value explicitly.

cyril-sf: 👍

mixonic:
Looks good! Great catch @cyril-sf, thank you :-)

PR #685, opened 2019-04-23
Move ember-decorators to devDeps

rwwagner90:


package.json
```diff
@@ -23,7 +23,6 @@
     "lint:files": "ember adde-lint --file-names"
   },
   "dependencies": {
-    "@ember-decorators/argument": "^0.8.16",
     "@ember-decorators/babel-transforms": "^0.1.1",
```
ezpuzz: what about this one?

rwwagner90: I think we need the babel transforms, since we were using classes still, but not sure.

addon/-private/column-tree.js
```diff
@@ -508,16 +508,19 @@ export default EmberObject.extend({
...
   },
 
   destroy() {
     this.token.cancel();
     get(this, 'root').destroy();
 
-    removeObserver(this, 'columns.@each.isFixed', this.sortColumnsByFixed.bind(this));
-    removeObserver(this, 'widthConstraint', this.ensureWidthConstraint.bind(this));
+    removeObserver(this, 'columns.@each.isFixed', this._sortColumnsByFixed);
+    removeObserver(this, 'widthConstraint', this._ensureWidthConstraint);
```
mixonic: This is a good fix, though it seems unrelated to the PR description 🤔

PR #686, opened 2019-04-24
Add back decorators/argument

rwwagner90:
It seems like we need this, but not sure why.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "lint:files": "ember adde-lint --file-names"
   },
   "dependencies": {
+    "@ember-decorators/argument": "^0.8.16",
     "@ember-decorators/babel-transforms": "^0.1.1",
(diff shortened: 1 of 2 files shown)


mixonic:
@rwwagner90 can you describe why you think it should be added back? Is there an error or something?

pzuraq:
Yeah I don't think this should be necessary, if you're encountering failures it's likely coming from somewhere else now that decorators aren't being included at all in `ember-table`

rwwagner90:
@mixonic @pzuraq when I use the `Addepar/master` branch I get failures on Ember canary with messages something like `computed decorators must be instances of a computed` or something. When I add this back and yarn link, those failures go away. I'll get the exact output for you later today.

rwwagner90:
We don't need this. Was a yarn caching issue.

PR #687, opened 2019-04-24
[BUG] Fix memory leaks related to vertical-collection

cyril-sf:
See html-next/vertical-collection#256


mixonic:
This is a SHA in `master` 👍

PR #688, opened 2019-04-24
Ensure rowMetas always contains truthy metaLike

rondale-sc:


addon/-private/collapse-tree.js
```diff
@@ -196,7 +196,7 @@ export const TableRowMeta = EmberObject.extend({
       selection.add(rowValue);
     }
 
-    let rowMetas = Array.from(selection).map(r => rowMetaCache.get(r));
+    let rowMetas = Array.from(selection).map(r => rowMetaCache.get(r)).filter(Boolean);
```
mixonic: Not sure how often this code is run, how many items are in the list or what the general budget for performance is...

But this one line of code 1) allocates a new array copying the contents of a `Set` in `from` 2) iterates that array and builds a new array w/ `map` 3) allocates a function for `map` as well 4) iterates and allocates *another* array with `filter`.

🙏 might we:

```suggestion
    let rowMetas = [];
    for (let r of selection) {
      let cache = rowMetaCache.get(r);
      if (cache) {
        rowMetas.push(cache);
      }
    }
```

I don't understand in what cases `selection` may not iterable and must be allocated into a new array. On line 60 above it is literally cast from an array into a `Set` 🤦‍♂️ sets are iterable, for/of will work just fine for a set.

bantic:
This should be fixed by #727, which has been released in `2.0.0-beta.9`

PR #691, opened 2019-05-22
Minimize CPs

mixonic:
For ember-tr and ember-td minimize CPs. This minimizes the watchers that need to be maintained.


mixonic:
I've had success with these changes in a downstream app. In a production build running:

```
console.time('scroll'); $('.ember-table')[0].scrollTop = 4000; window.requestAnimationFrame(() => window.requestAnimationFrame(() => console.timeEnd('scroll')));
```

The timing was ~750ms. After these changes it was ~100ms faster. This is not a very scientific number :-p

The idea is that most of the time spent during render is not in the rendering engine, but instead setup/teardown of dependencies. By creating less CPs we can minimize the necessary book-keeping. A bunch of work moves into the template layer, but that is already faster in old versions of Ember and improves throughout Ember's releases.

<details>
<summary>Most of the work is in `_sync`</summary>
Screen Shot 2019-05-23 at 9 08 25 AM
</details>

PR #692, opened 2019-05-24
[CHORE] Remove totally ember-decorators

cyril-sf:


mixonic:
@cyril-sf can you explain the "why" of this change in the PR description? Thank you.

Issue #509, opened 2018-04-23
Fixed columns cannot be resized


addepar-andy:
Currently all non scroll fixed columns can be resized.  Scroll fixed columns should be resizable as well.

pzuraq:
This should have been fixed by #521

PR #693, opened 2019-05-30
[PERF] Set default bufferSize to 1

cyril-sf:
By default, `ember-table` was allocating a buffer of 20 rows before and after the visible ones.

This default value forces a lot of computation in the case of tables with many columns and complex cells, from syncinc more data (from rows to cells for example) and triggering more observers
(based on the number of computed properties defined per cell for example).

`vertical-collection` also does not allocate the part of the buffer that should be above the visible first row on the initial render, as there is none when loading the table. This means that the initial scroll will also lead to the creation of 20 rows in the DOM (and the underlying computation to render the content).

From a pure UI/UX point of view, we don't need more than 1 row on each side of the table,
and it's already the default for `vertical-collection`.

### DIFF
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -125,7 +125,7 @@ export default Component.extend({
     @argument
     @type number
   */
-  bufferSize: defaultTo(20),
+  bufferSize: defaultTo(1),
 
   /**
     A flag that tells the table to render all of its rows at once.


mixonic:
I think this is obviously correct for Addepar right now, I'm curious if there is anyone else out there for whom this is notable.

cyril-sf:
@mixonic do you have concerns about merging this PR?

mixonic:
@cyril-sf I don't think we understand why the value is explicitly `20`. Perhaps @pzuraq can share thoughts.

* It is initially set in this commit: https://github.com/Addepar/ember-table/pull/521/commits/ea0dc6f5a05f04a8bbb2c5ff65deca6cca7b1907
* Which is this PR: https://github.com/Addepar/ember-table/pull/521
* PR might have had many motivations, but fixes https://github.com/Addepar/ember-table/issues/509
* At a high level it was maybe addressing something wrt column nesting, and implicitly with having a fixed column which can has the same resize logic, etc?

This is also easy to override in a consuming app by passing an argument to `ember-tbody`

cyril-sf:
While it's easy, it also means that `ember-table` doesn't deliver necessarily good performance out of the box.

20 is an arbitrary number, there's no reason for it to be the correct answer to any bug IMO.

rondale-sc:
Yeah, I wonder why VC's default is 1 though?  Seems strange, no?

pzuraq:
I believe we made this change because we were seeing a lot of lag if users scrolled quickly. While more items do get rendered, and more are rendered on initial render, performance should be equivalent for subsequent rerenders. That said, I don't have a strong preference here one way or another.

PR #695, opened 2019-06-06
[TEST] Add rowCount to page objects for thead, tbody and tfoot

cyril-sf:
In some tests, it is interesting to see if a row is properly inserted or removed.

Because `vertical-collection` tries to not render all the rows, we can't simply count the number of `tr`. Instead we use the `data-test-row-count` attribute to keep track of the number of available rows in a table.

### DIFF
--- a/addon-test-support/pages/-private/ember-table-body.js
+++ b/addon-test-support/pages/-private/ember-table-body.js
@@ -7,6 +7,13 @@ export default {
   scope: 'tbody',
 
   /**
+    Returns the number of rows in the body.
+  */
+  get rowCount() {
+    return Number(findElement(this).getAttribute('data-test-row-count'));
+  },
+
+  /**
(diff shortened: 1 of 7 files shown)


mixonic:
@cyril-sf This will be a performance knock if the `attributeBinding` is not stripped in production. I don't think we should ship this if we can't have it be a non-prod-only binding.

cyril-sf:
@mixonic I'll double check, but it should be stripped by `ember-test-selectors`

mixonic:
@cyril-sf what if a user of ember-table does not have `ember-test-selectors` installed? Then it would be present in prod, right? It seems like a poor experience that ember-table would require ember-test-selectors to provide its best production build.

cyril-sf:
I didn't push the correct code: `attributeBindings` is not present, so it doesn't impact production code of applications that are not using `ember-test-selectors`.

PR #697, opened 2019-06-10
[FEATURE] Text alignment can be defined per column

cyril-sf:
The `textAlign` property on a column definition accepts `left`, `center` and `right`.
When the property is set, a class is added to the cell (`ember-table__text-align-left`, `ember-table__text-align-center`, `ember-table__text-align-right`).

Also adds `ember-decorators/argument` as it's needed by the doc site.

### DIFF
--- a/addon/components/-private/base-table-cell.js
+++ b/addon/components/-private/base-table-cell.js
@@ -1,17 +1,31 @@
 import Component from '@ember/component';
 import { equal } from '@ember/object/computed';
-import { observer } from '@ember/object';
+import { observer, computed } from '@ember/object';
 import { scheduleOnce } from '@ember/runloop';
 
 export default Component.extend({
   // Provided by subclasses
   columnMeta: null,
+  columnValue: null,
 
-  classNameBindings: ['isFirstColumn', 'isFixedLeft', 'isFixedRight'],
+  classNameBindings: ['isFirstColumn', 'isFixedLeft', 'isFixedRight', 'textAlign'],
 
   isFirstColumn: equal('columnMeta.index', 0),
   isFixedLeft: equal('columnMeta.isFixed', 'left'),
   isFixedRight: equal('columnMeta.isFixed', 'right'),
+
+  /**
+   Indicates the text alignment of this cell
+  */
+  textAlign: computed('columnValue.textAlign', function() {
+    let textAlign = this.get('columnValue.textAlign');
+
+    if (['left', 'center', 'right'].includes(textAlign)) {
+      return `ember-table__text-align-${textAlign}`;
+    }
+
+    return null;
+  }),
 
   // eslint-disable-next-line
   scheduleUpdateStyles: observer(
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@addepar/prettier-config": "^1.0.0",
     "@addepar/sass-lint-config": "^2.0.1",
     "@addepar/style-toolbox": "~0.7.0",
+    "@ember-decorators/argument": "^0.8.21",
     "@types/ember": "^2.8.22",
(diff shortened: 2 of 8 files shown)


tests/dummy/app/pods/docs/guides/header/columns/template.md
```diff
@@ -184,3 +184,26 @@ reorder has occured.
 
   {{demo.snippet name='docs-example-resize-reorder-actions.hbs'}}
 {{/docs-demo}}
+
+## Text alignment
+
+A column can have its text aligned left, center or right by setting the `textAlign` property on the column definition.
+
+When the property is set, the cell will have the matching class (`text-align-left`, `text-align-center` or `text-align-center`).
```
twokul: While @mixonic rudely was looking over my shoulder, he proposed to namespace classes to avoid collisions, something to the extent of `ember-table__text-align-left`

package.json
```diff
@@ -48,6 +48,7 @@
     "@addepar/prettier-config": "^1.0.0",
     "@addepar/sass-lint-config": "^2.0.1",
     "@addepar/style-toolbox": "~0.7.0",
+    "@ember-decorators/argument": "^0.8.21",
```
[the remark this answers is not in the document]
mixonic: Ah this *is* a dev dep. Would still like some context for the addition though.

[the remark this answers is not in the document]
cyril-sf: The documentation relies on it.

cyril-sf:
@mixonic I can add it to the PR description, the explanation is in the commit message that updates `package.json` ( 4a824013dd8f39972f150c0454af5fdc86a8bdda )

```

File: tests/dummy/app/pods/docs/guides/header/columns/template.md (hunk 1 above)
Comment (cyril-sf) [reply tests/dummy/app/pods/docs/guides/header/columns/template.md]: Wouldn't that be something we do once consistently across the whole repo?

many classes do not have references to ember table (`is-sortable` for example) and some are prefixing with `et` (`et-header-resize-area`).

I chose `text-align-left` because it's what is used by ET1 ( https://github.com/Addepar/ember-table-legacy/blob/635316521c2f6e347a8b37fb75c972dd25f97bd5/addon/styles/addon.less#L63-L71), which allows for a simpler transition.

File: addon/components/-private/base-table-cell.js [hunk 3]
```diff
@@ -1,18 +1,32 @@
...
   isFixedLeft: equal('columnMeta.isFixed', 'left'),
   isFixedRight: equal('columnMeta.isFixed', 'right'),
 
+  /**
+   Indicates the text alignment of this cell
+  */
+  textAlign: computed('columnValue.textAlign', function() {
+    let textAlign = this.get('columnValue.textAlign');
+
+    if (['left', 'center', 'right'].indexOf(textAlign) >= 0) {
```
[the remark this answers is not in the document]
cyril-sf: 👍 

I was overheating and tried `contains` 😄

tests/dummy/app/pods/docs/guides/header/columns/template.md
mixonic: 🤷‍♂ This PR adds the styles. So this PR seems like a good place to not introduce a possible collision.

CSS class style compat with ET1 is a fine goal. Is that already something true? I thought the DOM changed so much that styles applied to ET1 are really not expected to work for ET2.

cyril-sf: It' not true for us as we don't customize based on the alignment. I don't know about other applications and how they scoped it (`.ember-table .text-align-center` would still work for example).

Happy to change as I don't think it's worth discussing that much, but it feels weird to add a 3rd style to name classes.

PR #706, opened 2019-07-10
DOCS: Fix rendering of addon docs pages

bantic:
* Upgrade ember-cli-addon-docs and related devDeps
 * Remove @ember-decorators/argument devDep

Convert custom-cell and custom-header docs components to classic.
Fixes issue where the table-customization addon docs page would throw
errors at runtime related to the `@argument` decorator.

 * Restructure docs index and application templates

Add a `nav.item` for 'docs.index', otherwise ember-cli-addon-docs
hits an assertion and throws an error in local development when it tries
and fails to associate the currentURL ('/docs') with a known docs route.

Remove some private-api `style=` component props and errant spacer `li`s in
the nav. In newer ember-cli-addon-docs its styles are namespaced (under
`docs-X` classes), and the styling of the nav no longer looked correct.

 * Upgrade angle-bracket polyfill

Upgrade the polyfill to allow using `<EmberTh::SortIndicator>`
nested angle-bracket invocation style on the table-customization docs page.
Prior to this upgrade, the page would throw errors when visited.


cyril-sf:
It feels like the CSS on the left menu is still off based on the screenshot. `Quickstart` is still too close to `INTRODUCTION` for example. Not sure if it's blocking.

bantic:
@cyril-sf You're right. I'm not sure how it happened but it looks like my screenshot was incorrect. This is a redo of the screenshot. The vertical spacing of the sidebar nav looks better here to me, do you agree?
image

tests/dummy/app/pods/docs/guides/main/table-customization/template.md
```diff
@@ -65,8 +65,8 @@ components:
           <h.row as |r|>
             <r.cell as |columnValue columnMeta|>
               <EmberTh::SortIndicator @columnMeta={{columnMeta}} />
-              Custom Header {{columnValue.name}}
-              <EmberTh::Resizehandle @columnMeta={{columnMeta}} />
+              {{columnValue.name}}
```
mixonic: By dropping `Custom Header` haven't you made this the same as the default header? 🤔

cyril-sf:
The part we need to understand is that we are losing the API guide by dropping off `ember-decorators`/classes.

tests/dummy/app/pods/docs/guides/main/table-customization/template.md
bantic: I guess that's true, but this demo is to show that you can customize whether (and where) the sort-indicator and resize-handle appear. I'm not sure why it said "Custom Header" in the first place, probably copy-pasta'd here from another demo on this page. I think it is fine to remove. (fwiw, this demo was added _after_ the last docs were published (ie v 2.0.0-beta.4), so it's never been seen "publicly").

bantic: In #708 I add some additional docs and interactivity around this demo (namely, the ability to turn on/off the sort-indicator and resize-handle), so I think when that is merged this demo will be fully comprehensible without the "Custom Header" text.

PR #713, opened 2019-07-11
[CHORE] Move ember-test-selectors from deps to devDeps

bantic:
ember-test-selectors does not need to be a dependency.
It was moved from devDeps to deps in #537, which appears to have been unintentional.

As an aside, the `data-test-*` attrs that ember-table uses, e.g. https://github.com/Addepar/ember-table/blob/4cbaee809bdb45874bc5e24f2d694210210866a7/addon/components/ember-table/component.js#L41
appear to leak out to consuming apps and should likely be removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "ember-getowner-polyfill": "^2.2.0",
     "ember-legacy-class-shim": "^1.0.0",
     "ember-raf-scheduler": "^0.1.0",
-    "ember-test-selectors": "^0.3.9",
     "ember-useragent": "^0.6.0",
     "hammerjs": "^2.0.8"
   },
@@ -86,6 +85,7 @@
     "ember-resolver": "^5.1.1",
     "ember-source": "~3.1.0",
     "ember-source-channel-url": "^1.0.1",
+    "ember-test-selectors": "^0.3.9",
     "ember-truth-helpers": "^2.0.0",
     "ember-try": "^1.1.0",
     "eslint-plugin-ember": "^6.2.0",


mixonic:
@bantic I think the idea *might* have been that `[data-test-` attributes are used in [the page object](https://github.com/Addepar/ember-table/pull/537/files#diff-4f9fa10c4c545a3103f33e58d2507303R84) of that PR which is in turn published for apps to use in test. Thus it was added as a dep so any consuming apps would strip the `data-test-`?

And additionally [where data-test- is passed an an arg to components](https://github.com/Addepar/ember-table/pull/537/files#diff-4f8c5d93ba7498fd9727c35d15b0b4a3R2) the author wanted those to be published as attributes?

I'm also unsure about if when data-test- selectors is installed in the app, but not in the addon's dependencies, if the stripping takes place on the `addon/` folder. Often build-time addons only operate on an `addon/` folder for the addon and the `app/` tree for an app.

Anyway, I'm fine moving this to a dev dependency or not 😬 it is just a complex situation. Can you document in README what happens for an end-user in a) dev/test b) prod when they 1) don't use ember-test-selectors 2) do use ember-test-selectors?

bantic:
@mixonic Ah, you're right. I thought I had tested this and confirmed that the `data-test-*` always leak to consumer apps but I just tried again and I see that they are stripped when running the consumer app in production environment. I had previously tried the published `2.0.0-beta.7` (not stripped) but the latest master (specifically, I pinned a local ember app to ember-table@aa56375ffae0d39325d8a) does strip them.
To add to the confusion, the `beta.7` is not even the `latest` tag on npm, `2.0.0-beta.4` is, so in the consumer app if you `ember install ember-table` you get that older beta version.

Anyway, I think the right step here is to *not* move test-selectors but *do* add some docs that the page object exists: #715

PR #709, opened 2019-07-10
DOCS: Add docs acceptance tests

bantic:
Builds on top of the fixes in #706 and should not be merged until after #706 lands.

[Diff of changes unique to this PR](https://github.com/Addepar/ember-table/compare/bantic:bantic/fix-docs-rendering-basic...bantic:bantic/docs/acceptance-tests).

Adds basic acceptance test that `/docs` route works and all the nav links render correctly.
Uses an env var `RUN_DOCS_TESTS` to only run these tests on the default version of ember.

PR #714, opened 2019-07-11
[DOCS]: Add basic acceptance test for docs/ routes

bantic:
Supersedes #709 — it took a bit more work to get the tests passing on CI so I closed that PR and opened this one.

Adds acceptance tests that `/docs` and all of the navigation items on the `/docs` page render without error.

In order to get these tests passing I had to make a few tangential changes:

### Change: Remove ember-auto-import and faker from `devDependencies`, add `ember-faker`

Without this change, the `/docs` acceptance tests fail with a message like:

```
Global error: Uncaught Error: Assertion Failed: You attempted to define a `{{link-to "docs.guides.header.sorting"}}` but did not pass the parameters required for generating its dynamic segments. Could not find module `faker` imported from `dummy/pods/docs/guides/header/sorting/controller`
```
[Failing Travis test run](https://travis-ci.org/Addepar/ember-table/jobs/556979019).

This appears to be due to Travis's CI build failing to import `faker`, which is used for some of the docs controllers.

`ember-auto-import`'s [Usage From Addons](https://github.com/ef4/ember-auto-import#usage-from-addons) docs state:

> ember-auto-import must be in the dependencies of your addon

and

> ember-auto-import will refuse to import devDependencies of your addon

This seems to explain why `faker` was not available for the CI test runs, as at the time it and `ember-auto-import` were both in `devDependencies`. I confirmed that moving them both to `dependencies` fixes CI, but seeing as how these are both used only for this addon's tests that did not seem an appropriate fix.

So instead I've removed them and brought back `ember-faker` in `devDependencies`. Unfortunately ember-faker only brings in faker@3 (current version is 4), so we move back to an earlier version of faker, but judging from [its changelog](https://github.com/Marak/faker.js/blob/master/CHANGELOG.md#v400) there is nothing particularly important we are missing. (faker@3 already provides faked bitcoin addresses; I'm not sure what else we need \s ).

### Change: Add code to `dummy/config/environment` to detect at build-time whether `ember-cli-addon-docs` is installed

This makes it so only ember-try scenarios with the docs addon installed will run the `/docs` tests. These tests don't need to be run in every ember-try configuration, only the one that we will publish from, which will presumably always be `master` (aka `ember-default` scenario).

In #709 I had done this using an env variable, but this made the travis.yml a bit messy. The run-time detection of `ember-cli-addon-docs` seems like a cleaner approach. When it doesn't see that the addon docs are present in `package.json`, those tests are `skip`'d so they still show up in the CI build output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "ember-a11y-testing": "^0.5.0",
     "ember-ajax": "^4.0.2",
     "ember-angle-bracket-invocation-polyfill": "^1.3.1",
-    "ember-auto-import": "^1.2.20",
     "ember-classy-page-object": "^0.5.0",
     "ember-cli": "~3.1.4",
     "ember-cli-addon-docs": "^0.6.13",
@@ -77,6 +76,7 @@
     "ember-decorators": "^2.2.0",
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.0",
+    "ember-faker": "^1.5.0",
     "ember-fetch": "^6.4.0",
     "ember-load-initializers": "^2.0.0",
     "ember-math-helpers": "^2.10.0",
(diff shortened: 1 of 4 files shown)


bantic:
@cyril-sf 	and @mixonic 	re-requesting your review since you looked at #709

Issue #726, opened 2019-07-17
unhandled exception for computing width in some applications


efx:
I've only observed this in our private application that has a number of different components and a complex data flow. I can debug this interactively as I think it would be more efficient than attempting a dummy reproduction. 

https://github.com/Addepar/ember-table/blob/74705ccde9da82fc210dafbcb57cea325eea0d18/addon/-private/column-tree.js#L631 throws an unhandled exception: Assertion Failed: Cannot call get with '_width' on an undefined object. While pausing execution, I also see it throw the same exception when entering this.root.width into the browser console.
The underlying error occurs because of this line https://github.com/Addepar/ember-table/blob/74705ccde9da82fc210dafbcb57cea325eea0d18/addon/-private/column-tree.js#L335 In that case, `meta` is `undefined`.

Both these PRs provide a fix:
#626 
#688 
I verified both versions of the patch in #688 avoid the exception.

bantic:
Thanks for this research. The description here was a helpful catalyst to make a bit of progress on this issue today.
I was able to create a minimal repro:
  * Render a table that has a parent row with many child rows, enough that they will be occluded (i.e., they are past the scroll overflow and do not render)
  * Click the checkbox on the parent row to select it and all its child rows
  * Without scrolling, un-check any child row

After that, this line https://github.com/Addepar/ember-table/blob/0fa3e3f08ce2781bb89bc1d74fdf052bc5bfadd3/addon/-private/collapse-tree.js#L205
will fail an assertion and throw an error with the message:
`Cannot call get with '_parentMeta._rowValue' on an undefined object.`

The error stems from the fact that when the child row's `TableRowMeta#select` method is called, all rows are part of the table's `selection`, but at this line:
https://github.com/Addepar/ember-table/blob/0fa3e3f08ce2781bb89bc1d74fdf052bc5bfadd3/addon/-private/collapse-tree.js#L199
[...]

PR #727, opened 2019-07-19
 Use objectAt to fetch rowValues (cause meta alloc)

mixonic:
A WIP to solve https://github.com/Addepar/ember-table/issues/726

Although the `selection` set contains raw values only, some other code (like the summing of selected counts for group selection state in `collapse-tree.js`) expect that all items in the `selection` set have a meta allocated.

The non-meta-allocated rows were added where the bare `children` property of a rowValue was being iterated. Instead with this change the `tree` is referenced for pulling out the children, meaning the meta cache system is exercised and metas allocated for any selected row.

Open questions:

* [x] Need a few tests. I think @bantic and I have some reasonable ideas about how to test this now.
* [x] What is the performance implication?
* [x] The old code was adding only children of a selected group to the explicitly selected group. It did so as it ascended the parent meta tree.
  * This implementation ends up iterating over already processed nodes (disregarded because the depth is unexpected, but still processed).
  * The old implementation I believe would be incorrect if the data is `{ name: 0, children: [{ name: 1, children: [{ name: 2 }, { name: 3 }] }, { name: 4, children: [{ name: 5 }] }] }`, `0` is selected, and you toggle `2` then `5` will never be explicitly selected.
  * A better implementation IMO would be to walk the parents until you find the final `isSelected` node in the parent tree, then walk tree nodes until the depth becomes `>=` the depth of that parent. That would then include all children of children.
* [x] We should review if there are any other uses of "raw" row values without going through the tree. Not every case is wrong, but they are all possibly not allocating a meta where one is needed.
* [x] Final hairbrained idea: The code that expects this meta to be allocated is itself running only a few lines below this. Is there an alternative bookkeeping which could be internal to this function and is somehow faster/better than allocating the meta? Probably not, but maybe.
  * [x] Regardless `selection` the Set is actually internal, and there is no reason we can't store an array in that with the meta for each value (or have a maintained-in-parallel Set), and that would avoid the several hits to the meta cache in this function.


bantic:
This looks good to me. I am working up a failing test case that we can use to ensure this fixes the specific scenario I described in https://github.com/Addepar/ember-table/issues/726#issuecomment-512993539

>  Regardless selection the Set is actually internal, and there is no reason we can't store an array in that with the meta for each value (or have a maintained-in-parallel Set), and that would avoid the several hits to the meta cache in this function.

I believe this is still covered, but it's worth keeping in mind that, in addition to a user interactively selecting rows, part of the ET API is that the `selection` property that's passed-in to the component can be modified (*without* user interaction) and ET needs to react appropriately to times when the selection is changed that way.

mixonic:
@bantic what I mean is that the actual variable, the `Set` of `selection` in this function, is [cast to an array before it is passed outside the function](https://github.com/Addepar/ember-table/blob/2c54a06c6345314d725a293c356f2097f6cfd20e/addon/-private/collapse-tree.js#L261), so we can consider the data in that set 100% internal. We simply need to ensure we process it into an array before the function passes it out.

bantic:
Added a failing test in #731 and I've confirmed that rebasing it onto this branch fixes that test.

I'm also going to take a stab at a test to cover this behavior that we think was buggy before:

> The old implementation I believe would be incorrect if the data is { name: 0, children: [{ name: 1, children: [{ name: 2 }, { name: 3 }] }, { name: 4, children: [{ name: 5 }] }] }, 0 is selected, and you toggle 2 then 5 will never be explicitly selected.

mixonic:
@bantic and I reviewed each of the checked off items above to solidify our understanding here. As followup he intends to:

* Document how `selection` values are passed in to `onSelect`. We believe the behavior is now consistent but it is a bit odd.
* Close the other issues attempting to resolve #726, and cut a beta/next
* Land #731, which has a test for the issue resolved in this PR.

There is also low-hanging performance work that could be done after this change. Our goal in this patch was just to make it work, but the implementation is a bit naive.

PR #730, opened 2019-07-19
[DOCS] Change styling for row-selection demo options

bantic:
Change the styling so the interactive row-selection demo
moves the radio button labels to the right of the button, and
add back in some alignment and spacing between the labels.
The docs were using a tailwind CSS class "pr-4" but tailwind is no longer
provided via addon-docs, so those styles were not being applied.

Before
image

After
image

### DIFF
--- a/tests/dummy/app/styles/tables.scss
+++ b/tests/dummy/app/styles/tables.scss
@@ -4,6 +4,19 @@
   }
 }
 
+.demo-options-group {
+  display: flex;
+
+  & *:first-child {
+    text-align: right;
+    flex-basis: 40%;
+  }
+
+  & > * + * {
+    padding-left: 2em;
+  }
+}
+
 .demo-container {
   position: relative;
   width: 100%;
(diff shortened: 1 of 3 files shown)


tests/dummy/app/styles/tables.scss
```diff
@@ -4,6 +4,19 @@
...
 
+.demo-options-group {
+  display: flex;
+
+  & *:first-child {
+    text-align: right;
+    flex-basis: 40%;
+  }
+
+  & > * + * {
```
mixonic: 😬 :mindblown:

mixonic: Given `<a><b /><c /></a>` this only matches `c`?

bantic: Haha, yeah. It selects "an element that follows an element". I discovered this recently and am probably over-using it... :) It's a cleaner alternative to a rule that sets some padding and then a ":first-child" or ":last-child" that removes the padding.

https://alistapart.com/article/axiomatic-css-and-lobotomized-owls/

PR #731, opened 2019-07-19
Test that unrendered (occluded) row selection can be managed

bantic:
This is a failing test for #726.

Also fixes a typo in an unrelated test assertion message.


bantic:
I just force-pushed now that #727 has landed, to confirm that the test in this PR is fixed by that code.

PR #746, opened 2019-07-29
[DOCS] Test clicking on snippets links on docs pages

bantic:
I noticed that one of the docs demos on the "Sorting" page was throwing an error when its 'component.js' snippet was clicked.

This adds a test to click all the snippets on all the docs pages, to keep them from going out of sync.


bantic:
You can see the issue at the current live deployed docs.

  * https://opensource.addepar.com/ember-table/docs/guides/header/fixed-columns#dynamic-fixed-columns
  * Click "component.js" snippet label for the last demo ("Dynamic Fixed Columns")
  * Error in console, no snippet shown

This PR fixes that, and a similar one on the "Sorting" docs page

PR #748, opened 2019-07-29
[BUG] Fix cases where manual selection includes unrendered or nonexistent rows

### LINKED ISSUE #747: bug after manually setting `selection` to include un-rendered rows
It's possible to trigger an error by:

  1. Manually `set` a `selection` for a table that includes a row that has not yet been rendered (it is occluded)
  2. Click a checkbox to modify the existing selection

The root cause is the same as that described by #727 and specifically in https://github.com/Addepar/ember-table/issues/726#issuecomment-512993539: When the `CollapseTree` maps `selection` to `rowMetas`...

https://github.com/Addepar/ember-table/blob/0fa3e3f08ce2781bb89bc1d74fdf052bc5bfadd3/addon/-private/collapse-tree.js#L199

...the unrendered row has no `rowMeta` yet, and as a result this line throws `Cannot call get with '_parentMeta._rowValue' on an undefined object.`:

https://github.com/Addepar/ember-table/blob/0fa3e3f08ce2781bb89bc1d74fdf052bc5bfadd3/addon/-private/collapse-tree.js#L205

bantic:
Modify `CollapseTree`:
  * Add a function `setupAllRowMeta` that walks all the table's rows and
sets up a row meta for each one. This is called lazily in the case where
the table has a `selection` that includes some rows that don't yet have
rowMetas. This can happen if the selection includes rows that haven't yet
rendered. The rowMeta for a row is lazily created when it is rendered,
so rows that haven't yet rendered won't have a row meta. This usually is
not a problem because in normal user interaction with a table, the user will
only interact with rows that are rendered. However, it's possible to
programmatically set a selection, and in this case a row in the selection
may not yet have been rendered.
  * Add `mapSelectionToMeta` to contain the extra complexity of finding the row metas (and calling of `setupAllRowMeta` if needed) for a selection

It's also possible for a `selection` to contain a row without a corresponding
rowMeta if that row is not part of the table's rows. This is an invalid selection,
and this commit adds an Ember.warning for times when we detect such a case.

Fixes #747

Modifies one of the 747 test cases to assert that the warning is triggered
when ET encounters a selection with a row that is not part of the table. The added warning has the id `ember-table.selection-invalid`.

Adds a `warn-handlers` test helper file that provides a function `registerTestWarnHandler` that can be used to capture and assert on runtime `Ember.warn` warnings that happen during a test.

### DIFF
--- a/tests/integration/components/selection-test.js
+++ b/tests/integration/components/selection-test.js
@@ -549,6 +549,7 @@ module('Integration | selection', () => {
         this.set('selection', rows.slice(-5));
 
         await table.rows.objectAt(0).checkbox.click();
+        assert.ok(table.rows.objectAt(0).isSelected, 'first row is selected');
         assert.ok(true, 'no error');
       });
 
--- a/tests/integration/components/selection-test.js
+++ b/tests/integration/components/selection-test.js
@@ -8,6 +8,7 @@ import { generateRows } from 'dummy/utils/generators';
 import { A as emberA } from '@ember/array';
 import { run } from '@ember/runloop';
 import { scrollTo } from 'ember-native-dom-helpers';
+import { registerTestWarnHandler } from '../../helpers/warn-handlers';
 
 let table = new TablePage({
   validateSelected(...selectedIndexes) {
@@ -577,6 +578,9 @@ module('Integration | selection', () => {
       });
 
       test('Issue 747: Programmatic selection that includes a row not part of `rows`', async function(assert) {
+        let capturedWarningIds = [];
+        registerTestWarnHandler((_message, { id }) => capturedWarningIds.push(id));
+
         let rows = generateRows(1, 1);
         await generateTable(this, { rows });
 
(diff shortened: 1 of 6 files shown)


tests/integration/components/selection-test.js
```diff
@@ -536,6 +537,67 @@ module('Integration | selection', () => {
...
+        await generateTable(this, { rows, bufferSize: 1 });
+
+        let renderedRowCount = table.rows.length;
+        assert.ok(renderedRowCount < 200, 'some rows are occluded');
+
+        // Select rows at the end that will not all have been rendered yet
+        this.set('selection', rows.slice(-5));
+
+        await table.rows.objectAt(0).checkbox.click();
+        assert.ok(true, 'no error');
```
mixonic: Should this also assert that the row is selected when you scroll to it?

bantic: I'll add an assertion that it is selected, but this row is at the top of the table so will already be in view.

PR #752, opened 2019-08-09
Added example of using CSS Flex with ember-table

octabode:
Per discussion on #topic-tables I've added my Twiddles for both a pure CSS version and a Bootstrap version.


bantic:
👏 👏 thank you!

Issue #759, opened 2019-09-06
IE11 polyfill issue


ctjhoa:
Hi,

We faced some issue with IE11.
In a new Ember app, without any specific configuration, in IE11, any table will not load.

![2019-09-06-150758_3837x1978_scrot](https://user-images.githubusercontent.com/1716173/64442056-ab6e3180-d0be-11e9-84a8-61b7f76b2496.png)

This is related to the `for ... of`  usage.

To work in IE11, users need to add this configuration in their `ember-cli-build.js` file:
```js
    'ember-cli-babel': {
      includePolyfill: true
    }
```

![2019-09-06-151722_1730x646_scrot](https://user-images.githubusercontent.com/1716173/64442823-12d8b100-d0c0-11e9-9c40-a4a59e7d83a8.png)

This is not documented and leads to complex installation flow.
In addition, there is no possibility to only include a polyfill set.

What I suggest, is to avoid the usage of `for ... of` which is not really needed and is transpiled to huge code block for a single loop.

https://babeljs.io/docs/en/babel-plugin-transform-for-of

Array's [forEach](https://api.emberjs.com/ember/3.8/classes/EmberArray/methods/forEach?anchor=forEach) should be used instead.

https://caniuse.com/#search=forEach

mixonic:
@ctjhoa it seems like we could just document the requirement that one should use `includePolyfill` for IE11 support? I'd prefer to go in that direction.

ctjhoa:
Enable `includePolyfill` is not without consequences for end users.
It force end users to include **all** polyfills to their project for **every browsers**.

I understand your decision from a project point of view but from a ember-table user this is costly for a browser which is still supported by ember (https://blog.emberjs.com/2018/02/14/ember-3-0-released.html)

mixonic:
Something is fishy here. For/of is syntax and would be transpiled (sometimes, but not always, a transpilation needs a polyfilled). Array#forEach is an API and would always be polyfilled.

Anyway I think replacing array iteration with a plain for loop would be fine.

(Quickly, from my phone)

ctjhoa:
@mixonic PR updated.

PR #760, opened 2019-09-07
Replace "for of" by classic for loop (#759)

ctjhoa:
#759

### DIFF
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -232,7 +232,8 @@ export const TableRowMeta = EmberObject.extend({
     if (selectingChildrenSelectsParent) {
       let groupingCounts = new Map();
 
-      for (let rowMeta of rowMetas) {
+      for (let i = 0; i < rowMetas.length; i++) {
+        let rowMeta = rowMetas[i];
         let parentRow = get(rowMeta, '_parentMeta._rowValue');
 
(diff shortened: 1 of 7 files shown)


ctjhoa:
@bantic not sure why CI failed as I do not modify templates

mixonic:
@ctjhoa I'd like to keep this using `for`, as that avoids a function allocation and execution.

addon/-private/collapse-tree.js
```diff
@@ -275,7 +277,9 @@ export const TableRowMeta = EmberObject.extend({
 function reduceSelectedRows(selection, groupingCounts, rowMetaCache) {
   let reducedGroupingCounts = new Map();
 
-  for (let [group, count] of groupingCounts.entries()) {
+  let groupingCountsEntries = Array.from(groupingCounts.entries());
```
mixonic: It is unfortunate to allocate an array for each of the maps.

But furthermore `entries` is a method on `Map`, and though IE11 has a `Map` implementation it doesn't have support for `entries`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/entries#Browser_compatibility

So there is still a need for a polyfil in IE11.

ctjhoa:
Ok I'm giving up on this one. I'll open another PR to document `includePolyfill`

Issue #764, opened 2019-09-11
Reordering drop-indicator needs to indicate left or right


frykten:
Hi!

Currently, the `.et-reorder-drop-indicator` element does not indicate whether the target column is positioned left or right of the under-reorder column---likewise what we can find on ember-light-table. It is very disturbing for users.

Would you consider thinking about this feature?

bantic:
closed by #766

PR #766, opened 2019-09-12
[FEAT] Add direction to reorder indicator

frykten:
#764

### DIFF
--- a/addon/-private/utils/reorder-indicators.js
+++ b/addon/-private/utils/reorder-indicators.js
@@ -33,6 +33,7 @@ class ReorderIndicator {
     let left = (elementLeft - containerLeft) * this.scale + scrollLeft;
     let width = elementWidth * this.scale;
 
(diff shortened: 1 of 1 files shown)


frykten:
I saw the build failed:

```Build failed.
Template Compiler Error (broccoli-persistent-filter:TemplateCompiler) in dummy/pods/docs/guides/body/occlusion/template.hbs:36:27
Closing tag `p` (on line 49) without an open tag.```
Seems like it's from an older merge.

bantic:
@frykten This looks great, thank you! It would also be helpful to update the docs to point out that these classes are now available. I'll open a separate issue to keep track of that

bantic:
I realized that, to avoid closing off future directional changes in the table (to support right-to-left languages), using class names like `before` and `after` would be more appropriate. No action necessary now; when/if we add support for rtl tables we will just need to make the change at that time.

PR #769, opened 2019-09-13
[DOCS] Add browser compatibility section

### LINKED ISSUE #759: IE11 polyfill issue

ctjhoa:
Fix #759

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,19 @@ You can also use the `ember-tfoot` component, which has the same API as
   {{/ember-table}}
 ```
 
+## Browser compatibility
+
+This project is written using an EcmaScript 6 javascript syntax.
+Babel doesn't include polyfill by default, so if you want to target legacy browsers (eg. IE11), 
+you need to add this to your `ember-cli-build.js`:
+
+```js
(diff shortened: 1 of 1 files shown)


bantic:
@ctjhoa thank you!

Issue #754, opened 2019-08-15
Tests fail on ember-beta


bantic:
Likely due to the recent release of ember 3.12, the ember-beta tests (which were ember-canary, and were failing) are now failing. The travis build matrix only allows canary to build, so the entire test suite is getting marked as failed.

Example recent ember-beta build: https://travis-ci.org/Addepar/ember-table/jobs/569853988

PR #772, opened 2019-10-23
Update ember-classy-page-object to latest ^0.6.0

bantic:


bantic:
Test failures appear to be legitimate, but unrelated to this change.
See https://github.com/Addepar/ember-table/issues/754 — the tests used to fail on beta (allowed to fail at CI), but in the meantime beta became 'release', and as a result CI now fails.

mixonic:
@bantic seems fine to merge and fix the build separately if you've evaluated the failures.

PR #773, opened 2019-10-25
Fix CI

### LINKED ISSUE #754: Tests fail on ember-beta

bantic:
Fixes #754 

Changes:
  * Remove legacy decorators usage from files in `tests/` (and drop the dep itself from package.json) — the majority of the diff
  * Several dependency upgrades, notably:
  * Use `ember-qunit` instead of deprecated `ember-cli-qunit`. This fixes the "cannot call `lookup` on undefined" error in render tests for the newer ember versions
  * Upgrade to ember-cli-babel v7. This fixed an issue where the class function properties (e.g., [this one](https://github.com/Addepar/ember-table/blob/44a7ae0237e327d305505091ecb9cd69c77eabf5/addon/-private/sticky/table-sticky-polyfill.js#L55-L57)) weren't parsed correctly when serving/building the addon — I think that previously one of the 2ndary deps from the decorators addon may have been adding this functionality
  * **Turn off** the column-reordering tests for ember release, beta, canary — due to a change in how observers work starting in Ember 3.13, reordering a column using synchronous observers became *very* slow. See #775


.eslintrc.js
```diff
@@ -1,14 +1,6 @@
...
   extends: ['@addepar', '@addepar/eslint-config/ember'],
   parser: 'babel-eslint',
-  parserOptions: {
-    ecmaFeatures: {
-      legacyDecorators: true,
-    },
-  },
-  env: {
-    es6: true,
-  },
```
bantic: babel-eslint's docs note that the ESLint official parser's `parserOptions` aren't necessarily supported when using this alternative parser. I found that these options didn't seem to be needed, so I removed them.
Docs: https://github.com/babel/babel-eslint#setup

bantic:
Added note in addon-docs and readme pointing out Ember 3.13 regressions

PR #777, opened 2019-10-29
Use async observer when target app is Ember 3.13+

### LINKED ISSUE #775: Reordering columns with Ember 3.13 is very slow (Ember regression?)
When using Ember-Table (latest: v2.1.4) with Ember v3.13.3, dragging-and-dropping a column to reorder takes a very long time to update when the column is dropped. And after the first reorder, additional attempts to reorder columns do not work properly: the drop indicator does not move from the end of the table.

It takes on the order of 1-2 seconds to handle the dropped column on a demo app I created with 5 columns and 20 rows. The more rows/columns, the worse the performance gets.

The issue is not present when using Ember 3.12.

The source of the slowdown is the `items.replace` call in Ember-Table's `utils/array.js`:

https://github.com/Addepar/ember-table/blob/a583ea7638b55c06bfbd57da63cbe5bf5ffab0d1/addon/-private/utils/array.js#L26

Digging deeper into Ember's internals, the 2 `notifyPropertyChange` calls from the ember-metal `arrayContentDidChange` function are particularly slow:

https://github.com/emberjs/ember.js/blob/133609c7766ef9c02060a07d8fd885f7d2f1940a/packages/%40ember/-internals/metal/lib/array_events.ts#L59-L63

Specifically, it seems like a lot of time is spent in the `activeObservers` loop of `flushSyncObservers`:

https://github.com/emberjs/ember.js/blob/70bcd9facdaf37ba19f60e6a10a511a34724f0f4/packages/%40ember/-internals/metal/lib/observer.ts#L208-L219

It seems like the timing of these may have changed due to the tracked-property changes that landed in Ember 3.13

Possibly related bugs:
  * https://github.com/emberjs/ember.js/issues/18225
  * https://github.com/emberjs/ember.js/issues/18417

## Repro

Demo repo is here: https://github.com/bantic/ember-table-3-13-demo

Link to a deployed (buggy) demo with Ember 3.13: https://et-3-13-bug.surge.sh
Link to a deployed (working) demo with Ember 3.12: https://et-3-12-no-bug.surge.sh

Note that the two deployed apps are built with ember env `development` to enable easier debugging.

Here's a gif showing the problematic behavior:

ember-table-3-13-reorder-bug

bantic:
Fixes #775 

This adds a util file `observer` that exports `addObserver`, `removeObserver` and `observer` functions that pass the `sync=false` when target app is Ember 3.13+.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,5 +8,23 @@ module.exports = {
     'ember/avoid-leaking-state-in-ember-objects': 'off',
 
     'ember-best-practices/require-dependent-keys': 'off',
+
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: '@ember/object',
+            importNames: ['observer'],
+            message: 'For compatibility, import observers from -private/utils/observer',
+          },
+          {
+            name: '@ember/object/observers',
+            importNames: ['addObserver', 'removeObserver'],
+            message: 'For compatibility, import observers from -private/utils/observer',
+          },
+        ],
+      },
+    ],
   },
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,12 +16,13 @@ module.exports = {
           {
             name: '@ember/object',
             importNames: ['observer'],
-            message: 'For compatibility, import observers from -private/utils/observer',
+            message: 'For compatibility, use `import { observer } from "-private/utils/observer"`',
           },
           {
             name: '@ember/object/observers',
             importNames: ['addObserver', 'removeObserver'],
-            message: 'For compatibility, import observers from -private/utils/observer',
+            message:
+              'For compatibility, use `import { addObserver, removeObserver } from "-private/utils/observer"`',
           },
         ],
       },
--- a/addon/-private/utils/observer.js
+++ b/addon/-private/utils/observer.js
@@ -37,12 +37,26 @@ function asyncAddObserver(...args) {
     method = args[3];
   }
 
-  return emberAddObserver(obj, target, path, method, sync);
+  return emberAddObserver(obj, path, target, method, sync);
 }
 
-function asyncRemoveObserver(key, target, method) {
+function asyncRemoveObserver(...args) {
+  let obj, path, target, method;
   let sync = false;
-  return emberRemoveObserver(key, target, method, sync);
+  obj = args[0];
+  path = args[1];
+  assert(
+    `Expected 3 or 4 args for addObserver, got ${args.length}`,
+    args.length === 3 || args.length === 4
+  );
+  if (args.length === 3) {
+    target = null;
+    method = args[2];
+  } else {
+    target = args[2];
+    method = args[3];
+  }
+  return emberRemoveObserver(obj, path, target, method, sync);
 }
 
 export const observer = USE_ASYNC_OBSERVERS ? asyncObserver : emberObserver;
--- a/addon/-private/utils/observer.js
+++ b/addon/-private/utils/observer.js
@@ -1,0 +1,50 @@
+import { gte } from 'ember-compatibility-helpers';
+import { assert } from '@ember/debug';
+
+// eslint-disable-next-line no-restricted-imports
+import { observer as eObserver } from '@ember/object';
+
+// eslint-disable-next-line no-restricted-imports
+import {
+  addObserver as eAddObserver,
+  removeObserver as eRemoveObserver,
+} from '@ember/object/observers';
+
+const USE_ASYNC_OBSERVERS = gte('3.13.0');
+
+function asyncObserver(...args) {
+  let fn = args.pop();
+  let dependentKeys = args;
+  let sync = false;
+
+  return eObserver({ dependentKeys, fn, sync });
+}
+
+function asyncAddObserver(...args) {
+  let obj, path, target, method;
+  let sync = false;
+  obj = args[0];
+  path = args[1];
+  assert(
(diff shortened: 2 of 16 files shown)


addon/-private/utils/observer.js
```diff
@@ -0,0 +1,50 @@
...
+} from '@ember/object/observers';
+
+const USE_ASYNC_OBSERVERS = gte('3.13.0');
+
+function asyncObserver(...args) {
+  let fn = args.pop();
+  let dependentKeys = args;
+  let sync = false;
+
+  return eObserver({ dependentKeys, fn, sync });
```
bantic: This appears to be an acceptable way to call `observer`, see

https://github.com/emberjs/ember.js/blob/133609c7766ef9c02060a07d8fd885f7d2f1940a/packages/%40ember/-internals/metal/lib/mixin.ts#L868
and 
https://github.com/emberjs/ember.js/blob/133609c7766ef9c02060a07d8fd885f7d2f1940a/packages/%40ember/-internals/metal/lib/mixin.ts#L840

It is not part of the docs yet, but I think that is simply because they haven't been updated.

addon/-private/utils/observer.js
```diff
@@ -0,0 +1,50 @@
+import { gte } from 'ember-compatibility-helpers';
+import { assert } from '@ember/debug';
+
+// eslint-disable-next-line no-restricted-imports
+import { observer as eObserver } from '@ember/object';
+
+// eslint-disable-next-line no-restricted-imports
+import {
+  addObserver as eAddObserver,
```
twokul: nitpick: could we rename `eAddObserver` to `emberAddObserver` maybe?

addon/-private/utils/observer.js
```diff
@@ -0,0 +1,50 @@
...
+  } else if (args.length === 4) {
+    target = args[2];
+    method = args[3];
+  }
+
+  return eAddObserver(obj, target, path, method, sync);
+}
+
+function asyncRemoveObserver(key, target, method) {
+  let sync = false;
```
twokul: could also just pass `false` to `eRemoveObserver` but it is definitely more declarative

.eslintrc.js
```diff
@@ -8,5 +8,23 @@ module.exports = {
...
     'ember-best-practices/require-dependent-keys': 'off',
+
+    'no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: '@ember/object',
+            importNames: ['observer'],
+            message: 'For compatibility, import observers from -private/utils/observer',
```
twokul: should it be "For compatibility reasons, use `import { addObserver, removeObserver } from '-private/utils/observer';`"?

bantic:
The `yarn.lock` change looks like it is because the wrong ember-source (3.13) was in the yarn.lock in #773 :(. The diff in this PR restores it to the right version.

addon/-private/utils/observer.js
```diff
@@ -0,0 +1,50 @@
...
+  );
+  if (args.length === 3) {
+    target = null;
+    method = args[2];
+  } else if (args.length === 4) {
+    target = args[2];
+    method = args[3];
+  }
+
+  return emberAddObserver(obj, target, path, method, sync);
```
bantic: I got the argument order wrong! 😱 
It should be `obj, path, target, method, sync`.
Luckily the tests caught it. :) Pushing a fix...

tests/integration/components/selection-test.js
```diff
@@ -585,7 +585,7 @@ module('Integration | selection', () => {
         let rows = generateRows(1, 1);
         await generateTable(this, { rows });
 
-        this.set('selection', [...rows, { fakeRow: true }]);
+        run(() => this.set('selection', [...rows, { fakeRow: true }]));
```
bantic: Adding the `run` around the `set` was required when using async observer (ie Ember 3.13+).
However in the collapse-tree tests below I had to use `await settled()` to get those tests to pass — wrapping the changes in a `run` was not sufficient.

mixonic: Wouldn't the end of a runloop still force all the async observers to flush?

https://github.com/emberjs/ember.js/blob/7d86e1242b9655f4995068ec2d189e46076517f5/packages/%40ember/runloop/index.js#L20

Or is the issue that the async observer's run `onEnd` and so if they schedule into a queue the existing runloop is not re-used?

Is that ideal?

bantic:
Chatted with @mixonic and @pzuraq on Discord — the use of `run` to ensure async observers have all flushed seems to work correctly in render tests, where the backburner queue should always be `flush`-ing just before render (see [here](https://github.com/BackburnerJS/backburner.js/blob/cfad13c120d0a0957b845d52fa190ac2ed632d30/lib/backburner/queue.ts#L105)). But in the collapse-tree tests where it doesn't use `render`, looks like we need to `await settled()` because when `flushAsyncObservers` schedules into 'actions' ([here](https://github.com/emberjs/ember.js/blob/7d86e1242b9655f4995068ec2d189e46076517f5/packages/%40ember/-internals/metal/lib/observer.ts#L182)) it may already be too late for the runloop (because the flush call is made [here](https://github.com/emberjs/ember.js/blob/7d86e1242b9655f4995068ec2d189e46076517f5/packages/%40ember/runloop/index.js#L20), when the runloop ends).

In light of that, it seems like the best option is to go forward with this change, but revisit whether Ember's internals need to change so that the async observers fire reliably no matter whether a runloop is in-progress.

The build is green so I'm going to go ahead and merge this.

Issue #780, opened 2019-11-05
Docs: Point out the new reorder-directional classes


bantic:
#766 added `et-reorder-direction-right` and `et-reorder-direction-left` classes when drag-dropping a column.

The docs should be updated to point out that these classes exist and can be used for customized styling.

bantic:
closed by #783

PR #783, opened 2019-11-06
[DOCS] Add new reorder-directional classes

frykten:
#780 

Also created a PR on Addepar/addepar-style-toolbox to add these new classes.


bantic:
👏 thanks!

Issue #787, opened 2019-11-20
Range selection does not work on first try


frykten:
Hello!

When using the ranged selection (shift key pressed), on the first try, nothing is selected.

When looking at the code, the culprit is quite obvious:
```
if (range) {
      // Use a set to avoid item duplication
      let { _lastSelectedIndex } = tree;

      let minIndex = Math.min(_lastSelectedIndex, rowIndex);
      let maxIndex = Math.max(_lastSelectedIndex, rowIndex);
```

Of course, `_lastSelectedIndex` is undefined therefore _Math.min/max_ is **NaN**.

PR #788, opened 2019-11-20
bugfix: first multiselection has no _lastSelectedIndex

frykten:
#787 

Should I add a test?

### DIFF
--- a/addon-test-support/pages/ember-table.js
+++ b/addon-test-support/pages/ember-table.js
@@ -66,13 +66,24 @@ export default PageObject.extend({
   },
 
   /**
-   * Selects a range of rows in the body
+   * Selects a range of rows in the body with simple click first
    *
    * @param {number} beginIndex
    * @param {number} endIndex
    */
-  async selectRange(beginIndex, endIndex) {
+  async selectRangeFromClick(beginIndex, endIndex) {
     await this.body.rows.objectAt(beginIndex).click();
+    await this.body.rows.objectAt(endIndex).clickWith({ shiftKey: true });
+  },
+
+  /**
+   * Selects a range of rows in the body with shift+click first
+   *
+   * @param {number} beginIndex
+   * @param {number} endIndex
+   */
+  async selectRangeFromShiftClick(beginIndex, endIndex) {
+    await this.body.rows.objectAt(beginIndex).clickWith({ shiftKey: true });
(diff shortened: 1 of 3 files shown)


mixonic:
@frykten Yes this should certainly get a test. I'm a bit surprised the functionality isn't already covered by a test. It is because the tested functionality uses click, then shift+click? And your issue is trigged by shift+click without a preceding click?

frykten:
@mixonic Yes, that's exactly it.

What would you advise: change the method _selectRange()_ used to select from a `normal click first => shift+click second` to `shift+click first => shift+click second` or since both should be available, let's test both?

mixonic:
Ah yes, that is the case: https://github.com/Addepar/ember-table/blob/8519064e2b7c6d596fb42ad7708e6caf6cf9c84e/addon-test-support/pages/ember-table.js#L74-L77 the test always uses a click as the starting input, then the second click is shift+click.

I think that means we have essentially left undefined what the behavior for shift+click when there is nothing selected should be.

Some existing UX to consider:

* In the MacOS finder shift+click when nothing is selected actually selects from item #1 to the clicked item.
* In a `<select multiple>` shift+click with nothing selected simply acts like a click.
* In a browser selecting text on a webpage shift+click with nothing selected acts like a click.
* In Google Sheets shift+click with nothing selected effectively select from the first item (the top-left cell) to the cell clicked.

Are there other cases to consider? We just need to make a design decision, I will raise it with some other maintainers as well.

mixonic:
IMO `selectRange` should stay as-is, and we should add tests for this specific case.

frykten:
@mixonic I don't see any other cases.
Could be interesting to chat about it, right. Gotta admit, we're more into `shift+click with nothing selected simply acts like a click`, here, but the other design is a good idea also.

I fixed-up the PR to add the test.

mixonic:
@frykten ok today I got to sync up with @bantic and @twokul and we think having a shift+click without any existing selection work as a click is fine. I think this PR already reflects that 👍 

@bantic looked at the CI quickly and it seems a point revision of prettier is causing the build to fail :-/

mixonic:
CI on master fixed as of https://github.com/Addepar/ember-table/pull/790, @frykten can you rebase this?

frykten:
Done @mixonic

mixonic:
Great! Thank you @frykten I'm not sure we will make a release for a week or two because of the holiday, but keep an eye out.

PR #790, opened 2019-11-21
Fix prettier lint errors

### LINKED ISSUE #789: prettier update breaks tests
Looks like tests are now failing with the "floating dependencies" scenario because prettier has updated and the linting is failing.

bantic:
Ran `prettier path/to/file --write` on the offending files.

Fixes #789.

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
   },
   "homepage": "https://Addepar.github.io/ember-table",
   "resolutions": {
-    "ember-cli-page-object": "1.15.1"
+    "ember-cli-page-object": "1.15.1",
+    "prettier": "1.18.2"
   }
 }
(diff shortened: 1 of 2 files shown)


mixonic:
@bantic I'm going to land this, as you are out for a week or so and this seems like a blocker.

...but I don't understand the change. The description here mentions running `write` over offending files, but no files besides dependencies are changed in this PR.

bantic:
> The description here mentions running write over offending files, but no files besides dependencies are changed in this PR.

Originally I did do what the description says, but it didn't fix things at CI — some of the changes I made to files (via prettier 1.19's `write` option) caused the lint tests to fail. I believe we were inadvertently using both new (1.19.x) and old (1.18.x) versions of `prettier` at CI, and fixing the errors reported by one caused failures from the other. [Here](https://travis-ci.org/Addepar/ember-table/jobs/615285738?utm_medium=notification&utm_source=github_status) is one of the intermittent failing CI runs. My hunch is that the autogenerated ESLint tests are somehow using older prettier and the best fix is to update whatever part (ember-cli, or ember-qunit ?) of the code generates those tests.

After I realized the initial attempt (running `write` over files) didn't fix CI I pushed the commit that pins pretter to 1.18 to see if that fixed the tests. My intention was to come back and make better sense of this before merging, if it did look like (as it does) prettier 1.19 and 1.18 weren't playing well together.

I'll add a follow-up issue to point out the discrepancies here.

PR #797, opened 2019-12-17
chore(deps): update ember-cli-sass to 10.0.1

bmish:
I had some Node 8 -> 10 upgrade issues because this dependency is out-of-date.


bmish:
According to the tests, this upgrade fails with some Ember 1.x and Ember 2.x versions, so it will need to be considered for a future major version release.

PR #799, opened 2020-01-03
remove .bowerrc

efx:
Continues https://github.com/Addepar/ember-table/pull/433
I'm not sure if we still need the configuration in
`config/ember-try.json`.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,4 +1,0 @@
-{
-  "directory": "bower_components",
-  "analytics": false
-}


efx:
>  The next major version of `ember-try` (v2.0) will drop support for installing dependencies with `bower`.
CI [answers](https://travis-ci.org/Addepar/ember-table/jobs/632392338#L817).

So I think we need to update ember-try and its configuration before removing this file.

PR #744, opened 2019-07-29
some backwards-compatible closure action usage (#614)

efx:
start to fix #614.
The way I safely check for the action feels repetitive so am open to
better approaches.
I noticed 2 forks of ember-table solve this so I think it makes sense to add backwards compatibility  into the addon. I followed Chris's [comment](https://github.com/Addepar/ember-table/issues/614#issuecomment-421419454).

PR #801, opened 2020-01-05
Deprecations/send action

billybonks:
Saw deprecations, so thought i would start to fix them. I created a util that should make it easy to replace in most places. After creating the pr i saw https://github.com/Addepar/ember-table/pull/744 with related work.


addon/-private/utils/send-action.js
```diff
@@ -0,0 +1,11 @@
+import { SUPPORTS_CLOSURE_ACTIONS } from 'ember-compatibility-helpers';
+
+export default function(context, action) {
+  let params = Array.from(arguments).slice(2);
+  if (SUPPORTS_CLOSURE_ACTIONS) {
+    if (typeof context[action] === 'function') {
+      return context[action](...params);
```
mixonic:
```suggestion
      context[action](...params);
```

addon/-private/utils/send-action.js
```diff
@@ -0,0 +1,11 @@
...
+    }
+  }
+  return context.sendAction(action, ...params);
```
mixonic:
```suggestion
  context.sendAction(action, ...params);
```

addon/-private/utils/send-action.js
```diff
@@ -0,0 +1,11 @@
+import { SUPPORTS_CLOSURE_ACTIONS } from 'ember-compatibility-helpers';
+
+export default function(context, action) {
+  let params = Array.from(arguments).slice(2);
```
mixonic: Why not simply `export default function(context, action, ...params) {`?

mixonic:
@billybonks can you additionally add a test which passes a closure action? The test would need to only run in versions of Ember which support that, of course.

addon/-private/utils/send-action.js
billybonks: Lol, just want sure with the syntax or what search terms I would use. So went with the techniques I remembered.

billybonks:
@mixonic @bantic i added an example of how this could be done. please reference the commit for an explanation https://github.com/Addepar/ember-table/pull/801/commits/ee6a45113eacb2ee9c99bf94a3a681bc6c742532

addon/-private/utils/send-action.js
```diff
@@ -0,0 +1,10 @@
+import { SUPPORTS_CLOSURE_ACTIONS } from 'ember-compatibility-helpers';
+
+export default function(context, action, ...params) {
+  if (SUPPORTS_CLOSURE_ACTIONS) {
+    if (typeof context[action] === 'function') {
+      context[action](...params);
```
fusion2004: Without the return statement here, couldn't this now trigger the action twice if it's a closure action?

mixonic: Yes. The proper factoring is:

```js
export default function(context, action, ...params) {
  if (SUPPORTS_CLOSURE_ACTIONS && typeof context[action] === 'function') {
    context[action](...params);
  } else {
    context.sendAction(action, ...params);
  }
}
```

The goal of not returning from the action is that we don't want to presume the return value from closure actions are available, since they won't be when calling `sendAction`.

tests/helpers/generate-table.js
```diff
@@ -10,7 +11,12 @@ import {
 // reexport for use in tests
 export { configureTableGeneration, resetTableGenerationConfig, generateColumns, generateRows };
 
-const fullTable = hbs`
+function actionString(actionName) {
+  let quotedActionName = `"${actionName}"`;
+  return SUPPORTS_CLOSURE_ACTIONS ? `(action ${quotedActionName})` : quotedActionName;
```
mixonic: Oh dang, this is clever but it is too much.

We need to test both versions of action explicitly, the string version and the closure action. We don't want to test only *one* of those options based on env.

I'd expect something like:

```js
if (!emberVersionGte('3.0')) { // is 3.0 right for this removal?
  test('test sendAction', /* */);
}
if (emberVersionGte('1.13')) {
  test('test (action', /* */);
}
```

...that way the 2.x range of Ember versions is testing both versions, and we don't need the runtime compiler and interpolation.

billybonks: so the only way to do to your recommendation without runtime compiler. would be to duplicate the `fullTable` string, since `onClick="onRowClick"` needs to become `onClick=(action "onRowClick")` or whichever variant. duplicating has a maintenance overhead that I wanted to avoid.

any idea how we can overcome that challenge?

addon/-private/utils/send-action.js
billybonks: great catch, made the change

tests/helpers/generate-table.js
mixonic: @billybonks I think that is the point- we need to have test coverage of both invocation styles. For example you could imagine you even want to test `onClick=this.actionViaActionDecorator` works. That could only be tested in 3.14(ish, whenever the decorator was added) though.

We could write our own babel plugin which converts based on the current version of Ember at build time:

```js
hbsWithClassicActionTranspilation`{{foo onClick=(action 'onRowClick')}}`

// can, based on a macro/transpilation step, be emitted as either:
hbs`{{foo onClick=(action 'onRowClick')}}`
hbs`{{foo onClick='onRowClick'}}`

// in fact in 3.14+ when the @action decorator is used you could also emit it as
hbs`{{foo onClick=this.onRowClick}}`
// thought I guess that requires a different declaration of the action in JS
// space, so maybe is out of scope for a compilation step.
```

efx:
Closed https://github.com/Addepar/ember-table/pull/744 in favor of this.

billybonks:
I'll try finish the request to change the tests today, it's half done at the moment

xomaczar:
Seems like this PR got stalled, what remaining? Can we push it over the finish line?

mixonic:
I flagged this as `3.0` since we should certainly resolve the concern, however I think the non-1.13 branch of https://github.com/Addepar/ember-table/pull/825 is probably closer to the final patch.

twokul:
Closing in favour of https://github.com/Addepar/ember-table/pull/860

PR #802, opened 2020-01-08
Avoid "update prop already used in computation"

### LINKED ISSUE #795: tests breaking with Ember v3.15
ember 3.15 was released yesterday (Dec 10), and so it is now the "release" version in the build matrix, causing the [build to fail](https://travis-ci.org/Addepar/ember-table/jobs/623723306):

```
not ok 90 Chrome 79.0 - [432 ms] - Integration | footer > basic: renders if footerRows are set
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: You attempted to update `rows` on `<(unknown):ember4815>`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
            
            `rows` was first used:
            
            - While rendering:
              ----------------
                application
                  index
                    ember-table
                      ember-tfoot
                        this.data-test-row-count
            
            Stack trace for the first usage: 
                at _get2 (http://localhost:7357/assets/vendor.js:34106:11)
                at http://localhost:7357/assets/vendor.js:21218:39
                at runInAutotrackingTransaction (http://localhost:7357/assets/vendor.js:33592:9)
                at track (http://localhost:7357/assets/vendor.js:33899:9)
                at RootPropertyReference.compute (http://localhost:7357/assets/vendor.js:21217:36)
                at RootPropertyReference.value (http://localhost:7357/assets/vendor.js:21104:43)
                at UpdateDynamicAttributeOpcode.evaluate (http://localhost:7357/assets/vendor.js:68329:36)
                at UpdatingVM.execute (http://localhost:7357/assets/vendor.js:71474:16)
            
            Stack trace for the update:
                at notifyPropertyChange (http://localhost:7357/assets/vendor.js:34919:7)
                at _set2 (http://localhost:7357/assets/vendor.js:35465:9)
                at Class.set (http://localhost:7357/assets/vendor.js:47734:29)
                at Class.<anonymous> (http://localhost:7357/assets/vendor.js:172431:25)
                at http://localhost:7357/assets/vendor.js:35976:36
                at untrack (http://localhost:7357/assets/vendor.js:33931:7)
                at ComputedProperty.get (http://localhost:7357/assets/vendor.js:35975:13)
        message: >
            Promise rejected during "renders if footerRows are set": Assertion Failed: You attempted to update `rows` on `<(unknown):ember4815>`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
```

mixonic:
Ember internals use an observer for attribute bindings. In this case the observer timing causes a computation based on a value which is then updated by a side-effect-having CP.

Here avoid the observer interaction by instead setting the count itself as part of the computation side effects.

See:

* https://github.com/Addepar/ember-table/issues/795
* https://github.com/emberjs/ember.js/issues/18613

Fixes #795

### DIFF
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -1,5 +1,6 @@
 import Component from '@ember/component';
 
+import { run } from '@ember/runloop';
 import { computed } from '@ember/object';
 import { observer } from '../../-private/utils/observer';
 import { bool, readOnly, or } from '@ember/object/computed';
@@ -234,7 +235,9 @@ export default Component.extend({
   */
   canSelect: bool('onSelect'),
 
-  'data-test-row-count': readOnly('wrappedRows.length'),
+  dataTestRowCount: null,
+
+  'data-test-row-count': readOnly('dataTestRowCount'),
 
   init() {
     this._super(...arguments);
@@ -297,7 +300,7 @@ export default Component.extend({
 
   /**
     Computed property which updates the CollapseTree and erases caches. This is
-    a computed for 1.11 compatibility, otherwise it would make sense to use
+    a computed for 1.12 compatibility, otherwise it would make sense to use
     lifecycle hooks instead.
   */
   wrappedRows: computed('rows', function() {
@@ -305,6 +308,11 @@ export default Component.extend({
 
     this.collapseTree.set('rowMetaCache', this.rowMetaCache);
     this.collapseTree.set('rows', rows);
+
+    run.schedule('actions', () => {
+      // eslint-disable-next-line ember-best-practices/no-side-effect-cp
+      this.set('dataTestRowCount', this.get('wrappedRows.length')); // eslint-disable-line ember/no-side-effects
+    });
 
     return this.collapseTree;
   }),
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,7 +66,7 @@ before_install:
   - export PATH=$HOME/.yarn/bin:$PATH
 
 install:
-  - yarn install --no-lockfile --non-interactive
+  - yarn install --no-lockfile --non-interactive --ignore-engines
 
 script:
   # Usually, it's ok to finish the test scenario without reverting


addon/components/ember-tbody/component.js
```diff
@@ -306,6 +309,11 @@ export default Component.extend({
     this.collapseTree.set('rowMetaCache', this.rowMetaCache);
     this.collapseTree.set('rows', rows);
 
+    run.schedule('actions', () => {
+      // eslint-disable-next-line ember-best-practices/no-side-effect-cp
+      this.set('dataTestRowCount', this.get('wrappedRows.length')); // eslint-disable-line ember/no-side-effects
```
twokul: do we need two `eslint-disable` comments here?

NicholasMohr: if this passes the test then LGTM but obviously this super sucks explicitly adding this sideeffectyness. Thankfully it's pretty limited side effect but still 🤢

mixonic: There are indeed two rules which error on this.

It is actually a bit annoying since the side effects really begin on line 309, but I presume the lint only catches on `this.set`

mixonic:
I believe the remaining failures are un-related to anything in ember-table, instead they should be resolved by https://github.com/ember-cli/ember-cli-htmlbars/pull/423 landing and getting released.

addon/components/ember-tbody/component.js
mixonic: Right, as noted this is expanding on an already-existing side effect which caused the problem. An alternative here would be to simply stop offering the `data-tests-row-count` attribute. If we drop 1.12 then this would be rewritten using `didReceiveAttrs` and the count could be set there without any CP side effects.

mixonic:
It is green! I had to add an additional commit to ignore engines wrt: https://github.com/npm/node-semver/commit/d61f828e64260a0a097f26210f5500e91a621828#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R35

This change makes the tests slightly more brittle but we don't need to stop testing with Node 8. Given that this addon nominally supports Node 8 that seems like the right tradeoff.

jiayingxu:
@mixonic with this change, the `rowCount` on the table body is not accurate for tree tables.  I created https://github.com/Addepar/ember-table/issues/804 to track this.

PR #803, opened 2020-01-10
[TESTS] Add tests for rowCount for tree tables

jiayingxu:
Add tests for `rowCount` for tree tables. When occlusion does not occur, the `rowCount` should match `rows.length`


mixonic:
Thanks for the reproduction

Issue #804, opened 2020-01-10
rowCount is broken for tree tables


jiayingxu:
After https://github.com/Addepar/ember-table/pull/802 landed, the `rowCount` on the table body is not accurate when using a tree table. 

I added test coverage for `rowCount` in tree tables in https://github.com/Addepar/ember-table/pull/803 which shows tests failing after this change
```
not ok 225 Chrome 79.0 - [201 ms] - Integration | Tree > basic: trees can be collapsed
    ---
        actual: >
            6
        expected: >
            4
        stack: >
                at Object._callee3$ (http://localhost:7357/assets/tests.js:7100:26)
                at tryCatch (http://localhost:7357/assets/vendor.js:6522:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:6748:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:6574:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:6981:105)
                at _next (http://localhost:7357/assets/tests.js:6983:196)
        message: >
            total number of rows removed
```

PR #805, opened 2020-01-10
Update test rowcount when collapse state changes

mixonic:
Fixes https://github.com/Addepar/ember-table/issues/804

Based on and replaces https://github.com/Addepar/ember-table/pull/803

The test rowcount did not update when collapse is triggered. Here do that, but via an observer in non-production builds only.

### DIFF
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -11,7 +11,7 @@ import CollapseTree, { SELECT_MODE } from '../../-private/collapse-tree';
 import defaultTo from '../../-private/utils/default-to';
 
 import layout from './template';
-import { assert } from '@ember/debug';
+import { assert, runInDebug } from '@ember/debug';
 
 /**
   The table body component. This component manages the main bulk of the rows of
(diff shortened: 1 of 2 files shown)


mixonic:
Still needs an approval!

mixonic:
> I cannot find usages of `data-test-row-count` in templates

When using ember-test-selectors the `data-test-` attribute is an attribute binding. So this is provided for test environments only.

I'm really not very happy with this fix, but it addresses the regression and I honestly don't think it is worse than the data-test- binding was before this change. I'm tempted to simply drop the test selector, but we shouldn't do it in a minor.

I've added `removeObserver`. The `collapseTree` object is destroyed just after the observers are removed and at first write I thought the `destroy` alone was sufficient for the teardown to be complete. I'm still looking into that but regardless we're gtg here.

mixonic:
> I cannot find usages of `data-test-row-count` in templates

When using ember-test-selectors the `data-test-` attribute is an attribute binding. So this is provided for test environments only.

I'm really not very happy with this fix, but it addresses the regression and I honestly don't think it is worse than the data-test- binding was before this change. I'm tempted to simply drop the test selector, but we shouldn't do it in a minor.

I've added `removeObserver`. The `collapseTree` object is destroyed just after the observers are removed and at first write I thought the `destroy` alone was sufficient for the teardown to be complete. I'm still looking into that but regardless we're gtg here.

Closing in favor of https://github.com/Addepar/ember-table/pull/806, ssh key hijinks.

PR #806, opened 2020-01-13
Update test rowcount when collapse state changes

mixonic:
Fixes https://github.com/Addepar/ember-table/issues/804

Based on and replaces https://github.com/Addepar/ember-table/pull/803

Replaces https://github.com/Addepar/ember-table/pull/805

The test rowcount did not update when collapse is triggered. Here do that, but via an observer in non-production builds only.


mixonic:
Merging based on prior approval via #805

PR #814, opened 2020-03-12
Add horizontal scroll indicator feature

kpfefferle:
Adds a new feature to optionally enable horizontal scroll indicator gradients on either side if there is scrollable overflow extending in that direction:

scroll-indicators

The scroll indicator gradients respect changes to the size of the ember-table container and table both, and they will render inside fixed columns if they are present. Includes documentation page demonstrating the feature and tests exercising the feature.

### DIFF
--- a/addon/components/-private/scroll-indicators/component.js
+++ b/addon/components/-private/scroll-indicators/component.js
@@ -60,7 +60,7 @@ export default Component.extend({
 
   _addListeners() {
     this._scrollElement = this._getScrollElement();
-    this._onScroll = this._updateIndicatorShow.bind(this);
+    this._onScroll = bind(this, this._updateIndicatorShow);
     this._scrollElement.addEventListener('scroll', this._onScroll);
     this._tableElement = this._getScrollElement().querySelector('table');
     this._resizeSensor = new ResizeSensor(
--- a/addon/components/-private/scroll-indicators/component.js
+++ b/addon/components/-private/scroll-indicators/component.js
@@ -62,7 +62,7 @@ export default Component.extend({
     this._scrollElement = this._getScrollElement();
     this._onScroll = bind(this, this._updateIndicatorShow);
     this._scrollElement.addEventListener('scroll', this._onScroll);
-    this._tableElement = this._getScrollElement().querySelector('table');
+    this._tableElement = this._scrollElement.querySelector('table');
     this._resizeSensor = new ResizeSensor(
       this._tableElement,
       bind(this, this._updateIndicatorShow)
--- a/addon/components/-private/scroll-indicators/component.js
+++ b/addon/components/-private/scroll-indicators/component.js
@@ -1,10 +1,10 @@
 import Component from '@ember/component';
 import { readOnly } from '@ember/object/computed';
+import { addObserver } from 'ember-table/-private/utils/observer';
 import layout from './template';
 
 export default Component.extend({
   layout,
-
   tagName: '',
 
   /**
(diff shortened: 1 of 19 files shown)


kpfefferle:
@aschenoni This may be the scroll indicator rendering _above_ the scroll bar. The scroll indicator element is a peer of the scrolling overflow container.

aschenoni:
I wrote a test and a fix. I'll submit a PR to this branch

aschenoni:
#821

addon/components/ember-table/template.hbs
```diff
@@ -1,8 +1,15 @@
-<table>
-  {{yield (hash
-    api=api
-    head=(component "ember-thead" api=api)
-    body=(component "ember-tbody" api=api)
-    foot=(component "ember-tfoot" api=api)
-  )}}
-</table>
+<div
```
mixonic, line 1: this is probably a semver breaking change as it alters the DOM?

addon/components/ember-thead/component.js
```diff
@@ -214,9 +221,10 @@ export default Component.extend({
     this._updateApi();
     this._updateColumnTree();
 
+    addObserver(this, 'enableScrollIndicators', this._updateApi);
+    addObserver(this, 'reorderFunction', this._updateApi);
```
mixonic, line 225: Is there anything significant about the reorder of observer initialization here?

addon/components/ember-thead/component.js
```diff
@@ -232,10 +240,11 @@ export default Component.extend({
 
   _updateApi() {
     this.set('unwrappedApi.columnTree', this.columnTree);
-    this.set('unwrappedApi.sorts', this.get('sorts'));
-    this.set('unwrappedApi.sortFunction', this.get('sortFunction'));
     this.set('unwrappedApi.compareFunction', this.get('compareFunction'));
+    this.set('unwrappedApi.enableScrollIndicators', this.get('enableScrollIndicators'));
```
mixonic, line 244: And the change in `set` order here?

addon/components/-private/scroll-indicators/component.js
```diff
@@ -0,0 +1,135 @@
...
+  columnTree: readOnly('api.columnTree'),
+  enableScrollIndicators: readOnly('api.enableScrollIndicators'),
+  tableScrollId: readOnly('api.tableId'),
+
+  leftStyle: indicatorStyle('left'),
+  rightStyle: indicatorStyle('right'),
+
+  _addListeners() {
+    this._scrollElement = this._getScrollElement();
+    this._onScroll = this._updateIndicatorShow.bind(this);
```
mixonic, line 67: I think this was intended to be an ember runloop `bind`. Generally we should just use arrow funcs and `run(` imo

addon/components/ember-table/template.hbs
kpfefferle: Yes, that thought crossed my mind, but positioning the scroll indicators on top of the table requires a common wrapping element 😬

addon/components/ember-thead/component.js
kpfefferle: I was just trying to alphabetize this group of observers rather than continue appending new ones after old ones.

addon/components/ember-thead/component.js
kpfefferle: Again, just alphabetizing the group.

addon/components/ember-table/template.hbs
kpfefferle: Retargeted to `3.0-beta` branch which should make this breaking change more palatable

addon/components/ember-table/component.js
```diff
@@ -91,7 +91,7 @@ export default Component.extend({
     return {
       columns: null,
       registerColumnTree: this.registerColumnTree.bind(this),
-      tableId: this.elementId,
+      tableId: `${this.elementId}-overflow`,
```
kpfefferle, line 94: @mixonic I believe that this update should keep row and column occlusion working as before. Yes, the scrolling element has changed, but I believe that this is the reference used for what element to observe the scroll on.

app/components/-ember-table-private/scroll-indicators.js
```diff
@@ -0,0 +1 @@
+export { default } from 'ember-table/components/-private/scroll-indicators/component';
```
twokul, line 1: why do we need this `export`?

addon/components/-private/scroll-indicators/component.js
```diff
@@ -0,0 +1,125 @@
...
+  tableScrollId: readOnly('api.tableId'),
+
+  leftStyle: indicatorStyle('left'),
+  rightStyle: indicatorStyle('right'),
+
+  _addListeners() {
+    this._scrollElement = this._getScrollElement();
+    this._onScroll = bind(this, this._updateIndicatorShow);
+    this._scrollElement.addEventListener('scroll', this._onScroll);
+    this._tableElement = this._getScrollElement().querySelector('table');
```
twokul, line 65: can we re-use `this._scrollElement` here?

addon/components/-private/scroll-indicators/component.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+  },
+
+  didInsertElement() {
+    this._super(...arguments);
+    this._updateIndicatorShow();
+    if (this.get('enableScrollIndicators')) {
+      this._addListeners();
+    }
+    addObserver(this, 'enableScrollIndicators', this._updateListeners);
```
twokul, line 117: do we need to set up the observer if `enableScrollIndicators` is `false`?

addon/components/-private/scroll-indicators/component.js
```diff
@@ -0,0 +1,125 @@
...
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { readOnly } from '@ember/object/computed';
+import { bind } from '@ember/runloop';
+import { htmlSafe } from '@ember/template';
+import { isEmpty } from '@ember/utils';
+import { addObserver } from 'ember-table/-private/utils/observer';
+import layout from './template';
+
+const indicatorStyle = side => {
```
twokul, line 11: I think a comment explaining what this function does would be 👌

twokul (commented):
Took a pass. Let me know if the comments make sense 😄

app/components/-ember-table-private/scroll-indicators.js
kpfefferle: Without this export, we can't render the `{{-ember-table-private/scroll-indicators}}` component in the `{{ember-table}}` template.

addon/components/-private/scroll-indicators/component.js
kpfefferle: Yes, because we need to add the listeners if the value of `enableScrollIndicators` is later changed to `true`

mixonic (approved):
Thanks for the excellent work and iteration on this @kpfefferle. It is a really exciting contribution, and addresses UX cases across many of our apps at Addepar.

I'll note that we've integrated this change into our (pretty large) test suites and have no regressions. We did need to adjust some tests for the DOM change.

Lets get it in and cut another beta.

PR #815, opened 2020-03-26
Use passed attributes to `EmberTable` for the table tag

ohcibi:
With the current layout one could not set a class to the `<table>` html element. By appending `...attributes` to that element with in this template one can 

```
<EmberTable class="table ...." title="An awesome table that supports native html attributes">
...
</EmberTable>
```

which would apply the class (and any other valid html attribute) to the table element.

### DIFF
--- a/addon/components/ember-table/template.hbs
+++ b/addon/components/ember-table/template.hbs
@@ -1,4 +1,4 @@
-<table>
+<table ...attributes>
   {{yield (hash
     api=api
     head=(component "ember-thead" api=api)


mixonic:
I believe to support `...attributes` an application must use the angle bracket invocation polyfill or be on at least Ember 3.4. As we will not ship the polyfill with this addon, and must continue to support versions prior to 3.4 in the near future, I'm going to close this.

Further feedback welcome if I have misunderstood the possibilities here.

PR #824, opened 2020-06-04
Add selectionMatchFunction option

### LINKED ISSUE #823: Programatic row selection seems broken?
I have an ember-table where I need to compute the selected row based off of the current route (because the table serves as the master in a master/detail view layout). When rendering a row's detail route, I want that row to be selected. I was doing this in a different part of the app successfully before, but that part of the app has moved away from that layout since, so I'm unsure when it stopped working.

Here's what I'm doing in the component that renders my table to compute the `selection` as the row which matches the currently rendered detail route (which renders in an `{{outlet}}` alongside the Orders ember-table:

```js
export default class OrdersTable extends Component {
  @service router;

  // ...

  get selection() {
    let { rows } = this;
    let { currentRoute, currentRouteName } = this.router;
    if (currentRouteName !== 'orders.order') {
      return undefined;
    }
    let orderId = currentRoute.params.orderId;
    return rows.find(row => row.id === orderId);
  }

  // ...

  @action
  onSelect(selection) {
    this.router.transitionTo('orders.order', selection.id);
  }
}
```

In the template, I pass `this.selection` and `this.onSelect` to `t.body`:

```hbs
<EmberTable as |t|>
  <t.head @columns={{this.columns}} />
  <t.body
    @checkboxSelectionMode="none"
    @onSelect={{this.onSelect}}
    @rows={{this.rows}}
    @rowSelectionMode="single"
    @selection={{this.selection}}
  />
</EmberTable>
```

Though `this.selection` returns the row object that I would expect, ember-table does not mark the matching table row with the `.is-selected` class 😬

kpfefferle:
This PR adds an optional `selectionMatchFunction` to the `ember-tbody` component. When set, this function will be used in place of `===` when checking to see if a row should be marked as selected.

Fixes #823 (at least offers a workaround)

### DIFF
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -64,16 +64,24 @@ export const TableRowMeta = EmberObject.extend({
     }
   ),
 
-  isGroupSelected: computed('_tree.selection.[]', '_parentMeta.isSelected', function() {
-    let rowValue = get(this, '_rowValue');
-    let selection = get(this, '_tree.selection');
+  isGroupSelected: computed(
+    '_tree.{selection.[],selectionMatchFunction}',
+    '_parentMeta.isSelected',
+    function() {
+      let rowValue = get(this, '_rowValue');
+      let selection = get(this, '_tree.selection');
+      let selectionMatchFunction = get(this, '_tree.selectionMatchFunction');
 
-    if (!selection || !isArray(selection)) {
-      return false;
+      if (!selection || !isArray(selection)) {
+        return false;
+      }
+
+      let isSelectionMatch = selectionMatchFunction
+        ? selection.filter(item => selectionMatchFunction(item, rowValue)).length > 0
+        : selection.includes(rowValue);
+      return isSelectionMatch || get(this, '_parentMeta.isGroupSelected');
     }
-
-    return selection.includes(rowValue) || get(this, '_parentMeta.isGroupSelected');
-  }),
+  ),
 
   canCollapse: computed(
     '_tree.{enableTree,enableCollapse}',
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -44,16 +44,25 @@ export const TableRowMeta = EmberObject.extend({
     },
   }),
 
-  isSelected: computed('_tree.selection.[]', '_parentMeta.isSelected', function() {
-    let rowValue = get(this, '_rowValue');
-    let selection = get(this, '_tree.selection');
+  // eslint-disable-next-line ember/use-brace-expansion
+  isSelected: computed(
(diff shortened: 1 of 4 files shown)


xomaczar:
@mixonic what's blocking this to be merged? missing tests?

xomaczar:
Never mind I’m not sure what this PR a fully does

twokul (commented):
👍 A couple of thoughts:

- can we add a test for this change?
- should we consider updating the docs to reflect this new API?

addon/components/ember-tbody/component.js
```diff
@@ -94,6 +94,14 @@ export default Component.extend({
   */
   selection: null,
 
+  /**
+    A function that will override how selection is compared to row value.
+
+    @argument selectionMatchFunction
+    @type function?
+  */
```
kpfefferle, line 102: This gets displayed in the docs for the `ember-tbody` component:

Docs

kpfefferle:
@twokul I added a test. For docs, see above how the JSDoc comment gets displayed on the docs website. I'm not sure that this function really warrants a full section in the examples.

twokul (approved):
Looks good to me!

Provisional ✅ (pending tests)

addon/-private/collapse-tree.js
```diff
@@ -44,27 +44,44 @@ export const TableRowMeta = EmberObject.extend({
...
-      return false;
+      let isSelectionMatch = selectionMatchFunction
+        ? selection.filter(item => selectionMatchFunction(item, rowValue)).length > 0
+        : selection.includes(rowValue);
+      return isSelectionMatch || get(this, '_parentMeta.isGroupSelected');
     }
-
-    return selection.includes(rowValue) || get(this, '_parentMeta.isGroupSelected');
-  }),
+  ),
```
kpfefferle, line 84: @twokul This set of changes is new and could probably use a final 👀

Adding the tests made me realize that the `isGroupSelected` CP *also* needed to use the `selectionMatchFunction` when matching `selection` against rows.

mixonic:
@kpfefferle did you get to follow up on the tests you wanted to add here?

kpfefferle:
@mixonic I have not yet done that and it fell off my radar. I'll try to carve out some time to get back to that this week.

addon/-private/collapse-tree.js
kpfefferle: Additional tests have been added to ensure that `selectionMatchFunction` works in both `single` and `multiple` modes

kpfefferle:
@mixonic @twokul Additional tests have been added to ensure we're covering both `single` and `multiple` selection modes. Ready for review!

twokul:
@kpfefferle Thanks! Merging this.

aemc:
an example would be nice for this func

PR #825, opened 2020-06-06
avoid sendAction when possible

### LINKED ISSUE #614: sendAction deprecation
Running ember-table with Ember v3.4 beta gave me the following deprecation:

```
DEPRECATION: You called <my-app@component:ember-tr::ember7583>.sendAction("onClick") but Component#sendAction is deprecated. Please use closure actions instead. [deprecation id: ember-component.send-action] See https://emberjs.com/deprecations/v3.x#toc_ember-component-send-action for more details.
```

juanazam:
Hello!

Been an ember-table user for quite some time, so I thought it was nice to give back to the community.

Closes #614 

Used addon suggested by @pzuraq on the issue.
Kept backwards compatibility for 1.12 users.

### DIFF
--- a/addon/components/-private/simple-checkbox.js
+++ b/addon/components/-private/simple-checkbox.js
@@ -1,5 +1,8 @@
 import Component from '@ember/component';
+import { get } from '@ember/object';
 import defaultTo from '../../-private/utils/default-to';
+
+import { gte } from 'ember-compatibility-helpers';
 
 export default Component.extend({
   tagName: 'input',
@@ -23,7 +26,13 @@ export default Component.extend({
   value: null,
 
   click(event) {
-    this.sendAction('onClick', event);
+    if (gte('1.13.0')) {
+      if (get(this, 'onClick')) {
+        get(this, 'onClick')(event);
+      }
+    } else {
+      this.sendAction('onClick', event);
+    }
   },
 
   change(event) {
@@ -36,6 +45,12 @@ export default Component.extend({
     this.element.checked = this.get('checked');
     this.element.indeterminate = this.get('indeterminate');
 
-    this.sendAction('onChange', checked, { value, indeterminate }, event);
+    if (gte('1.13.0')) {
+      if (get(this, 'onChange')) {
+        get(this, 'onChange')(checked, { value, indeterminate }, event);
+      }
+    } else {
+      this.sendAction('onChange', checked, { value, indeterminate }, event);
+    }
   },
 });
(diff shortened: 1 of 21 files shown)


mixonic:
The non-1.13 branch of this code is likely what should land into the `3.0-beta` branch before the final 3.0 is cut. I'm not sure this will end up as the PR, but flagged it `3.0` regardless since it should be addressed by then.

juanazam:
@mixonic do you want me to do any changes? I can help, just let me know, I can help with 3.0 in general as well

twokul:
@juanazam Thanks for the PR! If you have free cycles this week to work on it, that'd be great.

A few thoughts after looking at the code:

- I think we should target this PR against [3.0-beta](https://github.com/Addepar/ember-table/pull/831)
- We should not add `ember-compatibility-helpers` dependency (all the checks for 1.13+ should go away too, closure actions is the default behaviour)
- We still have methods that start with `sendAction` (`sendEventAction`, `sendFullAction`); we should likely rename them to something like `callClosureAction`
- We need to consider cases when a closure action is not defined (to prevent errors like `undefined is not a function`)
- There are a couple of cases when we pass a bound `sendAction` function ([here](https://github.com/Addepar/ember-table/blob/master/addon/components/ember-thead/component.js#L209) and [here](https://github.com/Addepar/ember-table/blob/master/addon/components/ember-tbody/component.js#L256)); we can no longer do that. We are going to have to be explicit about what type of actions both of those objects can call (ColumnTree, CollapseTree)

I'm happy to keep collaborating on this PR 😄

juanazam:
@twokul I understand all points listed but the last one

You mean renaming sendAction to something else? or something more complex than that?

Cheers!

twokul:
@juanazam I _think_ the following (I am not 100%):

```diff
this.columnTree = ColumnTree.create({
- sendAction: this.sendAction.bind(this),
+ onSelect: this.onSelect.bind(this),
+ onReorder: this.onReorder.bind(this),
  columnMetaCache: this.columnMetaCache,
  containerWidthAdjustment: this.containerWidthAdjustment,
});
```

Or

```diff
this.columnTree = ColumnTree.create({
- sendAction: this.sendAction.bind(this),
+ onSelect: () => this.onSelect?.(),
+ onReorder: () => this.onReorder?.(),
  columnMetaCache: this.columnMetaCache,
  containerWidthAdjustment: this.containerWidthAdjustment,
});
```

twokul:
@juanazam We _really_ appreciate your work on this PR but looking to make progress as fast as possible. I'm going to close this PR in favour of https://github.com/Addepar/ember-table/pull/860.

juanazam:
@twokul sounds good! suoer busy this week!

PR #829, opened 2020-08-07
Bump `ember-cli-sass` to `10`

twokul:
This PR does a few things:

- bumps `ember-cli-sass` to `10`
- changes node.js support to `12+`

Once and if this change lands, we should probably release a major version.


package.json
```diff
@@ -93,7 +93,7 @@
     "release-it": "^12.3.4"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
```
mixonic, line 96: This would require a major version bump re: semver

package.json
```diff
@@ -66,6 +65,7 @@
     "ember-cli-deploy-git-ci": "^1.0.1",
     "ember-cli-eslint": "^5.1.0",
     "ember-cli-inject-live-reload": "^2.0.1",
+    "ember-cli-sass": "^10.0.0",
```
mixonic, line 68: I think this should still be a dependency, why make it a dev dependency?

yarn.lock
```diff
@@ -10705,7 +10693,7 @@ node-sass-import-once@^1.2.0:
   dependencies:
     js-yaml "^3.2.7"
 
-node-sass@^4.6.0, node-sass@^4.7.2:
+node-sass@^4.6.0:
```
mixonic, line 10696: Bump `@addepar/style-toolbox` to `0.8.1` and this should also go away

mixonic:
And perhaps we also drop Ember < 2.4 from ember-try and then release the major. We could leave it all on a long-lived branch and do a pre-release

yarn.lock
twokul: we haven't published `0.8.1` yet

package.json
twokul: my mistake

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - '8'
+  - '12'
```
mixonic, line 6:
```suggestion
  - '10'
```

package.json
```diff
@@ -93,7 +93,7 @@
     "release-it": "^12.3.4"
   },
   "engines": {
-    "node": "8.* || >= 10.*"
+    "node": ">= 12.*"
```
mixonic, line 96:
```suggestion
    "node": ">= 10.*"
```

PR #830, opened 2020-08-07
3.0: Drop support for Ember pre-2.4

mixonic:
Additionally:

* Tweak the travis config to be more modern
* Ping a dependency of ember-cli-addon-docs to avoid a regression from a dependency of that library: https://github.com/ember-cli/ember-fetch/issues/547 (this could alternatively be fixed by upgrading ember-fetch to `^8.0.2` which likely means upgrading ember-cli-addon-docs)
* Drop some minor 1.13-compat code in testing

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,15 +14,16 @@
   "repository": "https://github.com/Addepar/ember-table",
   "scripts": {
     "build": "ember build",
-    "lint:js": "ember adde-lint --javascript",
-    "start": "ember serve",
-    "test": "ember test",
-    "test:all": "ember try:each",
-    "lint": "ember adde-lint",
-    "lint:sass": "ember adde-lint --sass",
+    "docs:deploy": "ember deploy production",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -73,9 +72,8 @@
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-export-application-global": "^2.0.0",
     "ember-faker": "^1.5.0",
-    "ember-fetch": "^6.4.0",
     "ember-load-initializers": "^2.0.0",
-    "ember-math-helpers": "^2.10.0",
+    "ember-math-helpers": "~2.11.3",
```
mixonic, line 76: This was required since `ember-math-helpers` bumps major version number of `ember-cli-htmlbars` in minor releases. To continue supporting Ember 2.x via bower (as we do in Ember try), 2.11.x is the last version of `ember-math-helpers` this codebase can depend on.

PR #837, opened 2020-10-27
[BUG] Fix positioning of table footer rows when rows are of varying heights

jiayingxu:
The `TableStickyPolyfill` is used to position header and footer rows. The individual `th` and `td` offset positions are calculated for the footer rows so they remain “sticky”. There is a bug in the logic for calculating the bottom offset for table footer rows where the `rows` were being iterated backwards while the `heights` were iterated forwards. This bug manifests when table footer rows have varying heights and the list of heights is not a palindrome. 

As shown in the image below, the table has 7 footer rows where the height of the 1st footer row is 74px and the height of all other footer rows is 26px. The bottom offset of the 6th row takes into account the height of the 1st row and not the 7th row due to the loop iteration direction bug.
The list `heights: [74, 26, 26, 26, 26, 26, 26]` is errantly being read in reverse order as `heights: [26, 26, 26, 26, 26, 26, 74]`. 
The ember-twiddle demonstrates the issue [here](https://ember-twiddle.com/35a7259172fd509c4da030ac803fa937
).

image

### DIFF
--- a/tests/unit/-private/table-sticky-polyfill-test.js
+++ b/tests/unit/-private/table-sticky-polyfill-test.js
@@ -52,14 +52,23 @@ const standardTemplate = hbs`
   </div>
 `;
 
-function constructMatrix(n, m, prefix = '') {
+/**
+ * Constructs a matrix m by n
+ * @param m Rows
+ * @param n Columns
+ * @param prefix Prefix string
+ * @param skipPrefixOnRowIndices Skip adding the prefix string to the row (m) indices specified in the list
+ * @returns {Array} matrix
+ */
+function constructMatrix(m, n, prefix = '', skipPrefixOnRowIndices = []) {
   let rows = emberA();
 
-  for (let i = 0; i < n; i++) {
+  for (let i = 0; i < m; i++) {
     let cols = emberA();
 
-    for (let j = 0; j < m; j++) {
-      cols.pushObject(`${m}${prefix}`);
+    for (let j = 0; j < n; j++) {
+      let skipPrefix = skipPrefixOnRowIndices.includes(i);
+      cols.pushObject(skipPrefix ? n : `${n}${prefix}`);
     }
 
     rows.pushObject(cols);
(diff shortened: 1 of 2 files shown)


bantic (approved):
Good catch, fix looks good! 👏 

Is there a way to confirm that the table header/footer rows do indeed span multiple lines, in the tests? Is it just the length of the prefix text that makes them take multiple lines? If the table were wider, would they cease being multiline rows?

Is it possible to, for instance, amend the tests with an assertion that the rows are indeed inequal heights?

Provisional ✔️ pending addressing ^.

tests/unit/-private/table-sticky-polyfill-test.js
```diff
@@ -100,6 +109,36 @@ function verifyFooter(assert) {
...
+/**
+ * Verifies multi line header when scrolled to the bottom of the table
+ * @param assert
+ */
+function verifyMultiLineHeader(assert) {
+  let firstCellRect = find('thead tr:first-child th:first-child').getBoundingClientRect();
+  let expectedOffset = firstCellRect.top;
+
+  findAll('thead > tr').forEach(row => {
+    let firstCellRect = row.firstElementChild.getBoundingClientRect();
```
bantic, line 121: Is `firstElementChild` a stable part of the DOM? Is there any better identifier (tag name, class name) that could be used to ensure that `firstCellRect` is what we think it is?

jiayingxu:
@bantic, good point about verifying that the table header/footer rows span multiple lines. Currently, it is dependent on the length of the prefix text and the width of the table. 

I've added a pre-condition assertion to verify that the header/footer rows have unequal heights.

tests/unit/-private/table-sticky-polyfill-test.js
jiayingxu: Good call out that `firstElementChild` is not supported across all modern browsers as it is unsupported in IE8, IE9, and Safari. I've modified the search string of `findAll` from `thead > tr` to `thead > tr > td:first-child` so we are directly searching for the first table cell of each header row and made the same modifications to the footer verification function as well.

PR #831, opened 2020-08-07
3.0 beta

twokul:
Test PR to make sure tests are good.

See: https://github.com/Addepar/ember-table/issues/819

PR #841, opened 2020-11-02
[tests] Fix "Floating Dependencies" tests

bantic:
Pin whatwg-fetch dependency to '3.3.1'.

This fixes an error `fetch is not defined - maybe your browser targets are not covering everything you need?`
that shows up when running the tests for the floating dependencies stage.

That error comes from ember-fetch, see ember-cli/ember-fetch#547

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
   },
   "homepage": "https://Addepar.github.io/ember-table",
   "resolutions": {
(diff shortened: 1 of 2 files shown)


mixonic:
Closing in favor of 3.0-beta https://github.com/Addepar/ember-table/pull/831

PR #842, opened 2020-11-02
[tests] Use Github Actions for tests

bantic:
**Note** This is based on the `3.0-beta` branch, not `master`.

Run tests at github actions. The open-source queue for the Travis tests is variable and sometimes the tests at travis are stuck in the queue for hours. Github Actions run much more quickly (and predictably).

Changes:
  * ✂️ .travis.yml file
  * Add .github/workflows/ci.yml
  * Add a resolution on ember-compatibility-helpers@1.2.1 in the ember-try scenarios for 2.4 and 2.8 to work around a bug in 1.2.2 when the `ember-source` package is not present, see https://github.com/pzuraq/ember-compatibility-helpers/issues/47

PR #844, opened 2020-11-17
[3.0] [tests] Only run "ember test" in primary workflow job

bantic:
**Note**: Targets base of `3.0-beta` not `master`

PR #842 accidentally runs *all* ember tests via `ember try:each` in the primary workflow job. This 
fixes that to only run the current dependencies. The matricized job runs the ember-try scenarios.


jiayingxu (approved):
LGTM

Issue #846, opened 2021-01-07
An old version of ember-test-selectors is busting depercation workflow: getWithDefault


NullVoxPopuli:
This is resolved in v5: https://github.com/simplabs/ember-test-selectors/blob/master/CHANGELOG.md

mixonic:
@NullVoxPopuli What does "busting deprecation workflow" mean?

NullVoxPopuli:
I should have added more details. oof.
My guess is either:
Emmer-cli-dendency-checker error or the deprecations hadn't yet been updated to the new format? (Deprecation staging stuff?)

NullVoxPopuli:
ok, now that I've seen this again, I can confirm:

folks using `throwOnUnhandled: true` _must_ silence `getWithDefault` deprecations, because ember-table uses and old version of ember-test-selectors

BobrImperator:
Seems like this could wait until 2.8 drop as per https://github.com/Addepar/ember-table/pull/884#issuecomment-805779479. or even longer actually, until it's decided to drop below 3.8 support - judging by provided changelog.

NullVoxPopuli:
@BobrImperator, we are using version `3.0.2-1`

BobrImperator:
Still looks like EmberTable 3.0 intents to drop support for Ember 2.8, while ember-test-selectors 5.0 require Ember 3.8+ if i'm not mistaken.

Please bear in mind, I'm not a maintainer, was just skimming through issues :D

References:
https://github.com/simplabs/ember-test-selectors/blob/master/CHANGELOG.md
https://github.com/Addepar/ember-table/releases/tag/v3.0.0

PR #847, opened 2021-01-07
upgrade ember-test-selectors to get around the getWithDefault deprecation

NullVoxPopuli:
See: https://github.com/Addepar/ember-table/issues/846

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,6 +44,7 @@
     "ember-compatibility-helpers": "^1.2.0",
     "ember-raf-scheduler": "^0.1.0",
     "ember-test-selectors": "^2.1.0",
+    "ember-test-selectors": "^5.0.0",
     "hammerjs": "^2.0.8"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -3027,6 +3027,13 @@ calculate-cache-key-for-tree@^1.0.0:
   dependencies:
     json-stable-stringify "^1.0.1"
 
+calculate-cache-key-for-tree@^2.0.0:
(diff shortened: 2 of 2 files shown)


mixonic:
@NullVoxPopuli can you rebase? I'd like to sanity check this against Addepar's internal usage but it seems reasonable.

NullVoxPopuli:
Rebased!

mixonic:
This was rebased onto 3.0.0-3 or something. I rebased it onto the `3.0-beta` branch which is what it targets and pushed.

mixonic:
@NullVoxPopuli per the changelog in https://github.com/simplabs/ember-test-selectors/blob/master/CHANGELOG.md, v2 of ember-test-selectors is the last version to support Ember versions before 2.16. As this library *must* support Ember 2.8 (or Addepar won't really be able to us it), we can't bump the version of the test selector dependency past 2.x at this time.

NullVoxPopuli:
Maybe we release another v2 version that addresses the deprecation?

mixonic:
@NullVoxPopuli Yeah honestly `getWithDefault` is very easy to avoid. It would be fairly straightforward to write the patch upgrading ember-test-selector 2.x if the maintainers are willing to accept it.

I will say that ember-table is not going to be able to drop support for 2.x (even 2.8) for a good while. I would plan for 6 months of Ember 2.8 support and a year of Ember 2.x support as a rough, very rough, estimation. I hope that context helps you figure out next steps for your own contributions/efforts.

NullVoxPopuli:
It does, thanks! 🎉

Issue #819, opened 2020-04-13
Drop support for IE11, Ember 1.12 (Major ver. RFC?)


mixonic:
Contributors keep stumbling on the legacy use of `sendAction` in this codebase (most recently at https://github.com/Addepar/ember-table/pull/818). Today we cannot remove `sendAction` since it is required for Ember 1.12 and this library supports Ember 1.12.

Additionally this library supported IE11 despite that browser being extremely uncommon in modern usage. Indeed at Addepar we no longer have a need to support IE11.

I propose that we drop support for these two platforms and release a major version of Ember Table. The task list would be something like:

* [ ] Drop Ember 1.12 from `ember-try.js` https://github.com/Addepar/ember-table/blob/master/config/ember-try.js#L13
* [ ] Drop usage of `sendAction`, instead call passed-in functions directly
* [ ] Drop support for IE11 in https://github.com/html-next/vertical-collection, release a major version (or whatever passes for it in that repo)
* [ ] Drop support for IE11 in this repo (not sure if there is anything to do beyond vertical collection's bump)
* [ ] Cut a major version of ember-table

Am I missing any other removals we might want to make as part of this step?

mixonic:
We've been discussing this with maintainers, and we intend to move forward on dropping IE11

efx:
see #800 for one conversation.

mixonic:
We're moving forward with this! I'm updating the issue description with more details on the execution and plan.

mixonic:
Ok! So we kicked off this goal in April 2020, and after a very long year we're nearly ready to pull the trigger on a stable 3.0 release. @twokul @kpfefferle and @ahmacleod have been landing some excellent work into the beta releases.

Additionally in that time Addepar has removed our own requirement that ember-table continue to support Ember 2.4. We're interested in upping the minimum supported version of Ember to 2.8.3.

Is there feedback from repo watchers on this step? If there is little in the way of feedback, we're inclined to bump the baseline and hope that helps our 3.x release cycle last a little bit longer.

twokul:
I would strongly support 2.8 😄

PR #848, opened 2021-01-07
Ember 2.4+ does not need the assign polyfill

NullVoxPopuli:


twokul:
Related to https://github.com/Addepar/ember-table/issues/819.

twokul:
I am not sure we want to merge this into `3.0-beta` yet since we would like to drop all the "< Ember 2.4" targets but keep Ember 2.4, for now. I believe, `assign` appears in Ember proper in 2.5.

mixonic:
I believe the correct fix here is to use native `Object.assign` and stop using Ember's API at all. Then we can drop the polyfill. The browser support commitment for ember-table is fuzzy, but we're certainly *not* supporting IE11 in 3.0 and that was really the blocker on using `Object.assign`.

https://caniuse.com/mdn-javascript_builtins_object_assign

mixonic:
The only use of `assign` in Ember Table *is* `Object.assign`. So, I'm really not even sure what this dependency is doing. Certainly on our supported browsers it is doing nothing. Merging this.

PR #849, opened 2021-01-08
Drop IE11 support

twokul:
Related to https://github.com/Addepar/ember-table/issues/819


mixonic:
I would approve if this was a PR!

mixonic (approved):

PR #853, opened 2021-01-13
Use scrollLeft to detect when scroll indicators should appear

ahmacleod:
If the table element has a negative `margin-left` style, the left scroll indicator will always show, even when scrolled all the way to the left. This negative margin trick is commonly used in conjunction with `border-spacing` and an overflow element in order to eliminate leading whitespace in the leftmost column of a table (`border-spacing` cannot be overridden at a cell level).

By examining `scrollLeft` instead of the bounding box positions of the table & overflow elements, the same effect is achieved without making any assumptions about the relative positions of elements.

### DIFF
--- a/addon/components/-private/scroll-indicators/component.js
+++ b/addon/components/-private/scroll-indicators/component.js
@@ -15,28 +15,24 @@ import layout from './template';
    @param {string} side - which side we are computing styles for: `left` or `right`
  */
 const indicatorStyle = side => {
-  return computed(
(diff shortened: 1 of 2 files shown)


kpfefferle (approved):
I tested this out in my consuming app, and it seems to work well. This seems like a more robust solution for the long-term since it accounts for this additional use case 👍🏻

ahmacleod:
Made redundant by https://github.com/Addepar/ember-table/pull/854

PR #854, opened 2021-01-19
Vertical scroll indicators

ahmacleod:
Vertical analogue of https://github.com/Addepar/ember-table/pull/814. Shows horizontal shadows when it is possible to scroll up or down. Respects any header or footer rows.

Extends https://github.com/Addepar/ember-table/pull/853

20210120 Vertical Scroll Indicators

With scrollbars-always-on:
20210122 Vertical Scroll Indicators Scrollbars Always On

Usage:
```
<EmberTable as |t|>
  <t.head
    @columns={{columns}}
    @scrollIndicators="all"
  />
  <t.body @rows={{rows}} />
</EmberTable>
```

Possible string values for `scrollIndicators` are `all`, `horizontal`, `vertical`, and `none`. Boolean values `true` and `false` are aliased to `all` and `none`, respectively.


twokul (approved):
Looks good to me! Could you add a gif to the PR description and maybe a code snippet (copy-paste from the one in the docs)?

Issue #771, opened 2019-09-19
Table body is visible behind table header


optikalefx:
On every header example on the demo page.

![image](https://user-images.githubusercontent.com/926885/65197478-356bb200-da50-11e9-849f-29021cde3a7c.png)

Mac OS 10.14.6. Chrome Version 76.0.3809.132

PR #859, opened 2021-01-28
Fix vertical scroll indicator placement

ahmacleod:
Addresses bottom scroll indicator placement when footer is large enough to scroll. Scroll indicator drops down to bottom of table when body is fully scrolled and obscured footer rows begin to scroll into view. Docs updated to show this example.

20210129 ET Footer Scrolling V2

Note the border rendering issues with the sticky footer cells. I double-checked, and this is existing behavior. Looks to be related to https://github.com/Addepar/ember-table/issues/771.

### DIFF
--- a/tests/dummy/app/pods/docs/guides/header/scroll-indicators/controller.js
+++ b/tests/dummy/app/pods/docs/guides/header/scroll-indicators/controller.js
@@ -63,6 +63,8 @@ export default Controller.extend({
   }),
 
   footerRows: computed(function() {
-    return generateRows(1);
+    return generateRows(100, 1, (row, key) => {
+      return String.fromCharCode(key.charCodeAt(0) + 7);
+    });
   }),
 });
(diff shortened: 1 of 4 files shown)


tests/dummy/app/pods/docs/guides/header/scroll-indicators/controller.js
```diff
@@ -63,6 +63,8 @@ export default Controller.extend({
   }),
 
   footerRows: computed(function() {
-    return generateRows(1);
+    return generateRows(100, 1, (row, key) => {
+      return String.fromCharCode(key.charCodeAt(0) + 7);
+    });
```
mixonic, line 68: By making this change here, have we lost test coverage for cases where the footer rows are *not* scrollable?

ahmacleod: _This_ change only affects the example in the docs. We could include an example for each case (single footer row vs. many scrollable), but this effectively demonstrates everything the bottom indicator does.

It's fair to ask the same question of the tests, however. I'll make sure both are covered explicitly.

ahmacleod: Done.

mixonic (approved):
Very cool.

PR #860, opened 2021-02-02
Remove `sendAction` API

twokul:
## Summary

As we are dropping support for older version of Ember.js, we need to make sure we trim the usage of old/deprecated APIs, as well. This time around it is the mechanism by which the actions are propagated, or `sendAction` API.

`sendAction` API has been [deprecated](https://github.com/emberjs/rfcs/blob/master/text/0335-deprecate-send-action.md) in favour of [closure actions](https://guides.emberjs.com/v2.4.0/components/triggering-changes-with-actions/) seems like forever ago (Ember 2.0+ but do not quote me on that.)

This PR makes sure we use *only* closure actions moving forward. It also contains a breaking change: the consumers that passed actions via a string (i.g. `onUpdateSorts="myAction"`) will have to change their code to `onUpdateSorts=(action "myAction")`.

This PR targets 3.0 release.

Closes https://github.com/Addepar/ember-table/pull/825
Closes https://github.com/Addepar/ember-table/pull/801
Related to https://github.com/Addepar/ember-table/issues/819

## Known Actions

- [x]  `onSelectionToggled`
- [x]  `onClick`
- [x]  `onCollapseToggled`
- [x]  `onUpdateSorts`
- [x]  `onReorder`
- [x]  `onResize`
- [x]  `onHeaderCellContextMenu`
- [x]  `onSelect`
- [x]  `onRowClick`
- [x]  `onRowDoubleClick`
- [x]  `onCellClick`
- [x]  `onCellDoubleClick`

### DIFF
--- a/addon/-private/collapse-tree.js
+++ b/addon/-private/collapse-tree.js
@@ -166,7 +166,7 @@ export const TableRowMeta = EmberObject.extend({
 
     if (single) {
       tree._lastSelectedIndex = null;
-      tree.sendAction('onSelect', rowValue);
+      tree.onSelect?.(rowValue);
       return;
     }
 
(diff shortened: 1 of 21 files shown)


mixonic (approved):
Epic

PR #865, opened 2021-02-04
Add "slack" width constraint modes

ahmacleod:
Adds two width constraint modes: `eq-container-slack`, `gte-container-slack`. These build on their non-"slack" counterparts by allocating excess container whitespace into an additional column that vanishes when it is no longer needed. This resembles behavior in applications like Excel and Google sheets.

**BONUS ADDITION:** `initialFillMode` parameter that is used exclusively in slack modes to perform the initial column layout. Values and default are the same as `fillMode`. In `eq-container-slack` mode, `fillMode` and `initialFillMode` can be set to different values. The former is used for enforcing the width constraint, while the latter is used only for initial sizing. See docs & tests for examples.

Which constraint modes use which fill modes:

**Advantages**:
- More closely resembles resize behavior in applications like Excel, Google Sheets
- Frees the rightmost column; e.g. if table has two columns and the right column is right-aligned, this makes it possible to bring the right column values to the left for easier comparison.
- Preserves the width of the rightmost column

**Notes**:
- The slack column cannot be reordered or manually resized
- There is no slack counterpart for `lte-container` or `none`
- `fillMode` has no effect in `gte-container-slack` mode, but _does_ still apply when enlarging columns beyond the container size in `eq-container-slack` mode
- `resizeMode` can be `fluid`, but `standard` (shown below) provides a more consistent UX
- `initialFillMode` should be used with slack modes if a "responsive" initial layout is desired

### eq-container-slack
20210204 eq-container-slack

### gte-container-slack
20210204 gte-container-slack

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,16 +41,18 @@ jobs:
       matrix:
         ember-version:
           [
-            ember-lts-2.4,
             ember-lts-2.8,
             ember-lts-2.12,
             ember-lts-2.18,
             ember-lts-3.4,
             ember-lts-3.8,
+            ember-lts-3.12,
             ember-lts-3.16,
+            ember-lts-3.20,
             ember-release,
             ember-beta,
             ember-canary,
+            ember-default-docs
           ]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,18 +41,16 @@ jobs:
       matrix:
         ember-version:
           [
+            ember-lts-2.4,
             ember-lts-2.8,
             ember-lts-2.12,
             ember-lts-2.18,
             ember-lts-3.4,
             ember-lts-3.8,
-            ember-lts-3.12,
             ember-lts-3.16,
-            ember-lts-3.20,
             ember-release,
             ember-beta,
             ember-canary,
-            ember-default-docs
           ]
 
     steps:
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -82,6 +82,8 @@ const TableColumnMeta = EmberObject.extend({
 
   isReorderable: readOnly('_node.isReorderable'),
 
+  isSlack: readOnly('_node.isSlack'),
+
   width: readOnly('_node.width'),
 
   offsetLeft: readOnly('_node.offsetLeft'),
@@ -224,6 +226,8 @@ const ColumnTreeNode = EmberObject.extend({
       let slackColumnNode = ColumnTreeNode.create({
         column: {
           isResizable: false,
+          isReorderable: false,
+          minWidth: 0,
         },
         tree,
         parent,
@@ -687,6 +691,9 @@ export default EmberObject.extend({
       }, 0);
 
       let slack = Math.max(containerWidth - contentWidth, 0);
+
+      // slack -= 8; // TODO: remove Glados-specific 8px adjustment
+
       let slackColumn = columns.findBy('isSlack', true);
 
       if (slack !== slackColumn.get('width')) {
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -36,13 +36,14 @@ export const FILL_MODE = {
   FIRST_COLUMN: 'first-column',
   LAST_COLUMN: 'last-column',
   NTH_COLUMN: 'nth-column',
-  SLACK_COLUMN: 'slack-column',
 };
 
 export const WIDTH_CONSTRAINT = {
   NONE: 'none',
   EQ_CONTAINER: 'eq-container',
+  EQ_CONTAINER_SLACK: 'eq-container-slack',
   GTE_CONTAINER: 'gte-container',
+  GTE_CONTAINER_SLACK: 'gte-container-slack',
   LTE_CONTAINER: 'lte-container',
 };
 
@@ -205,7 +206,7 @@ const ColumnTreeNode = EmberObject.extend({
     }
   },
 
-  subcolumnNodes: computed('column.subcolumns.[]', 'tree.fillMode', function() {
+  subcolumnNodes: computed('column.subcolumns.[]', 'tree.widthConstraint', function() {
     this.cleanSubcolumnNodes();
 
     if (get(this, 'isLeaf')) {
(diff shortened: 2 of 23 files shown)


tests/dummy/app/pods/docs/guides/header/size-constraints/template.md
```diff
@@ -1,20 +1,20 @@
...
 1. `eq-container`: Ensures that the table is always exactly the width the of its
 container.
 
-2. `eq-container-slack`: Similar to `eq-container`, but allocates excess whitespace to an empty slack column on the right side of the table.
+2. `eq-container-slack`: Similar to `eq-container`, but allocates excess whitespace to an empty "slack" column on the right side of the table.
 
 3. `gte-container`: Ensures that the table is always the same width or larger than its container.
 
-4. `eq-container-slack`: Similar to `gte-container`, but allocates excess whitespace to an empty slack column on the right side of the table.
+4. `eq-container-slack`: Similar to `gte-container`, but allocates excess whitespace to an empty "slack" column on the right side of the table.
```
kpfefferle, line 13: I assume this isn't named the exact same as option 2?

```suggestion
4. `gte-container-slack`: Similar to `gte-container`, but allocates excess whitespace to an empty "slack" column on the right side of the table.
```

ahmacleod: Good catch.

addon/-private/column-tree.js
```diff
@@ -214,6 +220,28 @@ const ColumnTreeNode = EmberObject.extend({
       get(this, 'column.subcolumns').map(column => ColumnTreeNode.create({ column, tree, parent }))
     );
 
+    let widthConstraint = tree.get('widthConstraint');
+
+    if (
```
twokul, line 225: It might be worth pulling this into a CP? What do you think?

```javascript
isSlackModeEnabled: computed('widthConstraint', function() {
  let widthConstraint = tree.get('widthConstraint');
  return widthConstraint === WIDTH_CONSTRAINT.EQ_CONTAINER_SLACK || widthConstraint === WIDTH_CONSTRAINT.GTE_CONTAINER_SLACK;
});
```

addon/-private/column-tree.js
```diff
@@ -624,42 +653,68 @@ export default EmberObject.extend({
     this._isSorting = false;
   },
 
-  ensureWidthConstraint() {
+  ensureWidthConstraint(isInitialRun = false) {
```
twokul, line 656: It is worth explaining what `isInitialRun` does 🤔 ✍️

addon/components/ember-thead/component.js
```diff
@@ -333,11 +347,11 @@ export default Component.extend({
     this.onUpdateSorts?.(newSorts);
   },
 
-  fillupHandler() {
+  fillupHandler(isInitialRun = false) {
```
twokul, line 350: It seems like we always call `fillupHandler` w/ `true` argument. Is that right?

twokul (approved):
Noted a few things. LGTM otherwise! 🎉

addon/-private/column-tree.js
ahmacleod: I like it.

addon/-private/column-tree.js
ahmacleod: Definitely.

addon/components/ember-thead/component.js
ahmacleod: The table resize sensor calls it with a `false` argument, but I'll double check if it's actually necessary to make the distinction.

ahmacleod:
Refactored to get rid of `isInitialRun` entirely. We now have `performInitialLayout()` and `ensureWidthConstraint()`. No funny business.

addon/-private/column-tree.js
ahmacleod: Refactored away.

addon/components/ember-thead/component.js
ahmacleod: Eliminated argument in favor of a distinct `performInitialLayout()` method called on init.

.github/workflows/ci.yml
```diff
@@ -41,18 +41,16 @@ jobs:
       matrix:
         ember-version:
           [
+            ember-lts-2.4,
```
twokul, line 44: I am positive we want to drop 2.4 from our support matrix ;)

addon/-private/column-tree.js
twokul, line 225: Probably can use `isSlackModeEnabled` CP here?

addon/-private/column-tree.js
```diff
@@ -624,39 +667,78 @@ export default EmberObject.extend({
     this._isSorting = false;
   },
 
+  performInitialLayout() {
```
twokul, line 670: I think we should add a comment

addon/-private/column-tree.js
```diff
@@ -624,39 +667,78 @@ export default EmberObject.extend({
...
+    let isSlackModeEnabled = get(this, 'isSlackModeEnabled');
+
+    if (isSlackModeEnabled) {
+      this.updateSlackColumn();
+    }
+
+    this.applyFillMode();
+  },
+
+  updateSlackColumn() {
```
twokul, line 699: I think we should add a comment

config/ember-try.js
```diff
@@ -10,6 +9,37 @@ module.exports = function() {
     return {
       useYarn: true,
       scenarios: [
+        {
+          name: 'ember-lts-2.4',
```
twokul, line 13: I am positive we want to drop 2.4 from our support matrix ;)

twokul (changes requested):
The code has changed since my last review. Noted a few things ✍️

config/ember-try.js
ahmacleod: Temporary change, gone now.

.github/workflows/ci.yml
ahmacleod: Temporary change, gone now.

addon/-private/column-tree.js
ahmacleod: Yes, much better.

addon/-private/column-tree.js
ahmacleod: Done

addon/-private/column-tree.js
ahmacleod: Done

addon/-private/column-tree.js
ahmacleod: Done

PR #869, opened 2021-02-06
Bump ember-classy-page-object

mixonic:
Bump https://github.com/Addepar/ember-classy-page-object which will also result in https://github.com/san650/ember-cli-page-object getting bumped.

TODO:

* [x] ~See if we have a ton of deprecations from ember-cli-page-object now~ There definitely are a bunch of deprecations, but it doesn't seem like we should block this PR on them. Many of them need to be addressed by updating the component integration test suite, which we need to assert can be done with Ember 2.8. IMO we should decouple fixing deprecations from landing the dependency bump.
* [x] See if @ro0gr can follow through on a new release of ember-cli-page-object fixing an Ember-triggered deprecation wrt `deprecate` usage (ohmy)
* [x] Get a release of ember-classy-page-object to depend on. Currently @twokul holds the keys to the kingdom.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "@html-next/vertical-collection": "2.0.0",
     "broccoli-string-replace": "^0.1.2",
     "css-element-queries": "^0.4.0",
-    "ember-classy-page-object": "^0.6.1",
+    "ember-classy-page-object": "^0.7.0",
     "ember-cli-babel": "^7.12.0",
     "ember-cli-htmlbars": "^3.0.1",
     "ember-cli-htmlbars-inline-precompile": "^2.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "@html-next/vertical-collection": "^1.0.0-beta.14",
     "broccoli-string-replace": "^0.1.2",
     "css-element-queries": "^0.4.0",
-    "ember-classy-page-object": "^0.6.1",
+    "ember-classy-page-object": "Addepar/ember-classy-page-object#714cb4523f7cbab01e7539b0ddfe7d81714ba919",
```
mixonic, line 37: Can't be landed until there is a release here.

ro0gr:
> See if @ro0gr can follow through on a new release of ember-cli-page-object fixing an Ember-triggered deprecation wrt deprecate usage (ohmy)

just published new version https://github.com/san650/ember-cli-page-object/releases/tag/v1.17.6

mixonic:
Thank you very much @ro0gr! I appreciate the fast turn around! :sweat_smile: You are a champ.

mixonic:
This branch passed an internal smoke test on Addepar's main Ember codebase 👍 I'll land it and cut a release. It should address the deprecation message noted at https://github.com/Addepar/ember-table/pull/831#issuecomment-774114559

PR #873, opened 2021-02-11
Improve footer tracking for bottom scroll indicator

ahmacleod:
In order to position the bottom scroll indicator correctly, we need to know if and when the visible portion of the footer changes height. Presently, we detect when the dimensions of the table's container changes, but not specifically when the height of the footer, or the position of its sticky cells, change. These changes can occur without altering the overall height or width of the table, which can leave the bottom scroll indicator floating in the wrong position.

So far this has only been a problem in testing environments, where it manifests as an erratic race condition during initial render. It is exacerbated by increased functionality in the footer (e.g. child rows), because of increased jostling during layout. During a typical application pageload, there are sufficiently many layout changes that the effect is rarely seen.

Since it is possible to programmatically add and remove a footer from the table, we install and remove those listeners dynamically.

Changes:
- Dynamically installs `ResizeSensor` on footer to detect changes in total footer height
- Dynamically installs `MutationObserver` on footer to detect changes to the `style` attribute of `td` elements; we are interested in when `bottom` changes value

### DIFF
--- a/addon/components/-private/scroll-indicators/component.js
+++ b/addon/components/-private/scroll-indicators/component.js
@@ -2,7 +2,7 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
 import { readOnly } from '@ember/object/computed';
-import { scheduleOnce } from '@ember/runloop';
+import { bind, scheduleOnce } from '@ember/runloop';
 import { capitalize } from '@ember/string';
 import { htmlSafe } from '@ember/template';
 import { isEmpty, isNone } from '@ember/utils';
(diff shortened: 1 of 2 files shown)


addon/components/-private/scroll-indicators/component.js
```diff
@@ -269,6 +317,12 @@ export default Component.extend({
 
   didInsertElement() {
     this._super(...arguments);
+
+    // debounced, runloop-safe version
+    this._updateIndicators = bind(this, () => {
```
twokul, line 322: Why do we need `bind` here?

twokul (commented):
The code change looks fine. Could you explain your approach?

addon/components/-private/scroll-indicators/component.js
ahmacleod: I think that's a leftover, let's nix it.

ahmacleod: Aha, I forgot this is why:
image

You need to wrap the `scheduleOnce` in a run loop or else the test suite gets mad. You can't get at it from the test itself.

twokul: Two more questions on this topic:

- Why do we set `this._updateIndicators` inside `didInsertElement` vs. `init`?
- Are there any issues with using an anonymous function vs. a function defined on the component?

twokul (approved):
Left a comment. Provisional ✅

addon/components/-private/scroll-indicators/component.js
ahmacleod: 1) No objection; consider it moved.

2) Not afaik. Anything we move to a component function is ultimately going to have to be wrapped in a `bind` or something that produces an anonymous function anyway. This forward declaration does a lot of heavy lifting:

* Context binding for a scroll listener, two `ResizeSensor`s, and a `MutationObserver`
* Deals with the test runloop situation
* Encapsulates the `scheduleOnce`
* Simplifies removing the scroll listener on teardown; don't have to cache a `this._scrollListener` function

What we _can_ do is dump the `bind` for a `run`. The context is already available via the anonymous closure, so `run` is a more precise solution to its problem and can be documented as such.

ahmacleod: `run` won't cut it for `ember-2.8-lts` test suite. I've notated the seemingly redundant `bind` for now; we can visit when we drop 2.8 support.

PR #874, opened 2021-02-25
Add columns array to onReorder and onResize

johanrd:
Enables storing column settings upon change (e.g. persisting to local storage)

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -973,7 +973,7 @@ export default EmberObject.extend({
 
     this.container.classList.remove('is-reordering');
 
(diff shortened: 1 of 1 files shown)


mixonic:
@johanrd hopefully quick questions:

* Are the columns and that array already public API?
* Where should this change be documented? Can you do so?

johanrd:
Hi. When you ask, I now see that I can pass it as part of the public api like this:

```
<EmberTable as |tb|>
  <tb.head
    @onResize={{fn this.onResize tb.api.columnTree.columns}}
  />  
</EmberTable>  
  ```

I guess we could make the documentation of[ EmberTable:api](https://opensource.addepar.com/ember-table/docs/api/components/ember-table#api) more clear.

Thanks! closing.

PR #875, opened 2021-02-25
Upgrade ember-cli-htmlbars and remove ember-cli-htmlbars-inline-precompile

johanrd:
Triggered by deprecation message in ember-cli
```console
DEPRECATION: ember-table > ember-cli-htmlbars-inline-precompile is no longer needed with ember-cli-htmlbars versions 4.0.0 and higher, please remove it from `/Users/johanroed/fremby/proapi-webapp/UI/console/node_modules/ember-table/package.json`
```


johanrd:
Closing, as I see that the need to keep supporting ember 2.8 and node 6 might keep us back here.

PR #876, opened 2021-03-02
Add is-last-column and is-slack classes to table cells

ahmacleod:
Changes:
- Adds an `is-slack` class to slack `th` and `td` elements
- Adds an `is-last-column` class to the rightmost column cells. In slack modes, the rightmost column can be either the slack column (when slack is visible) or its immediate neighbor to the left (when no slack is visible).
- Adds tests for the above and the previously-untested `is-first-column` class
- Improves the page object to better handle slack and nested columns

### DIFF
--- a/addon-test-support/pages/-private/ember-table-body.js
+++ b/addon-test-support/pages/-private/ember-table-body.js
@@ -9,6 +9,16 @@ import { findElement } from 'ember-classy-page-object/extend';
 
 import { click } from 'ember-native-dom-helpers';
 
+export const BodyCell = PageObject.extend({
+  scope: 'td',
+
+  doubleClick: triggerable('dblclick'),
+
+  isFirstColumn: hasClass('is-first-column'),
+  isLastColumn: hasClass('is-last-column'),
+  isSlack: hasClass('is-slack'),
+});
+
 export default PageObject.extend({
   scope: 'tbody',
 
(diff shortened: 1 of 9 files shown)


addon-test-support/pages/-private/ember-table-header.js
```diff
@@ -131,14 +139,14 @@ export default {
...
+   * Selects all non-slack `th` header elements from all header rows.
    */
   headers: collection('th:not([data-test-ember-table-slack])', Header),
 
   /**
-   * List of columns in the header, excluding slack column if present.
+   * Selects all slack `th` header elements from all header rows.
    */
-  contentHeaders: collection('th:not([data-test-ember-table-slack])', Header),
+  slackHeaders: collection('th[data-test-ember-table-slack]', Header),
```
twokul, line 149: I am assuming this private API. Is that correct?

twokul (approved):
Looks good. Left a few notes ✍️

addon-test-support/pages/-private/ember-table-header.js
ahmacleod: Good question. Moved "offline" for a broader discussion.

In this case, `contentHeaders` is redundant, unused internally, and new since 3.0. It's probably safe to assume nobody is using it. More clarity would be good here though.

addon-test-support/pages/-private/ember-table-body.js
```diff
@@ -9,6 +9,16 @@ import { findElement } from 'ember-classy-page-object/extend';
 
 import { click } from 'ember-native-dom-helpers';
 
+export const BodyCell = PageObject.extend({
```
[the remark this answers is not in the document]
ahmacleod: Done.

Issue #820, opened 2020-04-21
reorder not working


sharath-sriram:
Using ember-cli 3.14, ember-table 2.2.2, column reordering does not work - as in, when the column is reordered, the rows do not change according to the new position of the columns. However, subsequent reorder operations after the first one work properly.

Before reorder: 
<img width="570" alt="Screenshot 2020-04-21 at 5 54 17 PM" src="https://user-images.githubusercontent.com/59554914/79865726-3fe09680-83f9-11ea-8c5f-80732b3f1b63.png">

After reorder (The contact column is pushed to the end)
<img width="572" alt="Screenshot 2020-04-21 at 5 54 55 PM" src="https://user-images.githubusercontent.com/59554914/79865758-4f5fdf80-83f9-11ea-9374-36188a5ce247.png">

As it can be noticed in the screenshots, even though the Contacts column has moved, the rows themselves have not moved. Am I doing something wrong? Is it an internal bug?

sanmaverick:
I am observing this issue as well. Anyone else?

lfcorrea:
It is happening with me too. The reorder is not working on Ember versions newer than 3.12 :(

Any news about it?

EduardoLima806:
I am facing that same issue. Any guess estimate to solve it in the next release ?

Valgueiro:
Any updates about this?

sanmaverick:
Looking at the silence here, its likely that people have found a workaround? If there is a workaround for this, can someone please list it here?

tempo22:
Hi,

I have been having the same issue. It is related to the `dynamicAlias` computed, I still haven't figured why but it seem that this computed is processed after the template

<img src="https://user-images.githubusercontent.com/1882437/95178510-d0e8d580-07bf-11eb-956e-ab9e3e1abfe3.png" width="400">

I made a quick fix on a fork but I'm sure there is a deeper problem in there:

```
diff --git a/addon/components/-private/row-wrapper.js b/addon/components/-private/row-wrapper.js
index 83c2221..f47b315 100644
--- a/addon/components/-private/row-wrapper.js
+++ b/addon/components/-private/row-wrapper.js
@@ -8,7 +8,9 @@ import { objectAt } from '../../-private/utils/array';
 import { dynamicAlias } from '../../-private/utils/computed';
 
 const CellWrapper = EmberObject.extend({
-  cellValue: dynamicAlias('rowValue', 'columnValue.valuePath'),
[...]

Issue #808, opened 2020-01-21
Row cellValue not updating when columnValue.valuePath is updated (Ember >= 3.13.0)


napafundi:
When using new versions of Ember (>= 3.13.0) ember-table isn't updating `row.cellValue` correctly even though `columnValue.valuePath` has been changed. This occurs when setting a new array of columns to the `table.head.columns` property.

Here is a simple Ember project displaying this issue:
[ember-table-issue](https://github.com/napafundi/ember-table-issue)

billdami:
I can also confirm that this issue exists still even with the just released Ember v3.16.0 and v3.17.0-beta.1. Its worth noting that the first version of ember that this issue started occurring, 3.13, was the Octane "preview release", and contained a large amount of changes to Computed Properties, introducting tracked properties, etc. So it seems likely that this large internal changes to the object model have something to do with this bug.

For anyone else that encounters this, a very crude workaround can be implemented by forcing ember-table to recognize a change to the `rows` objects immediately after changing columns,  e.g. passing in a new temp array, and then changing it back to the original rows array in the next runloop:

```js
@tracked isChangingColumns = false;

@action
changeColumns() {
     //do something that changes `this.columns`...
    
    this.isChangingColumns = true;
    next(this, () => this.isChangingColumns = false);
}
```
```hbs
<T.body @rows={{if this.isChangingColumns (array) this.rows}} />
```

PR #878, opened 2021-03-09
fix: Ensure column cell values update when columns are reordered

HeroicEric:
Fixes https://github.com/Addepar/ember-table/issues/808
Fixes https://github.com/Addepar/ember-table/issues/820

The idea behind this is basically the same as https://github.com/Addepar/ember-table/pull/567.

The value of `cellValue` is computed by grabbing the `columnValue.valuePath`
and then grabbing the value of the corresponding key on `rowValue`.
Basically something like `rowValue.${columnValue.valuePath}`, which is
not possible using any of the built in computed macros.

This basically uses an observer to watch for changes to
`columnValue.valuePath`, and defines/redefines `cellValue` as a computed
alias for the property at that path on `rowValue`. This ensures the
`cellValue` is correct if `columnValue.valuePath` or the actual value at
that path on `rowValue` changes. Observers aren't recommended but this
was already using one.

Previously this was solved by creating a more generic `dynamicAlias`
computed macro in https://github.com/Addepar/ember-table/pull/567. To be
honest, I was having a little trouble wrapping my head around all that
was happening in there but I think the changes in this PR accomplish the
same idea.

I'm not totally sure what the issue was with the other implementation
but, since the `dynamicAlias` macro was only used in this one place, it
feels OK to have a more simple implementation that is hard coded
specifically for this case.

--- 

Side note: I was also able to get this working by replacing the `CellWrapper` with a tracked version of the same class, which allowed me to avoid using any observers. I'd be interested in exploring some more to see if it would be feasible to conditionally swap out some of the pieces of this addon to use `@tracked` if that's something the maintainers would be open to.


mixonic (approved):
This looks great @HeroicEric. I agree that it seems a massive simplification. If I can imagine a downside, it would be that there is a performance characteristic to the custom computed alias which is desirable.

That said I'm inclined to default toward correctness regardless unless there is a failure at the scale of "occlusion no longer works" or something.

Approving and will nudge for another maintainer to review as well.

ahmacleod (approved):
That's the _only_ place that entire module is used? I'm in.

mixonic:
This is now released in `3.0.2-1`, a beta release. Please test and let us know if there are any bad behaviors!

MichalBryxi:
Is there a reason to not bump the official release (or whatever is the name of the default version one gets from npm)? Currently it's `v3.0.1`, which does not have this patch and I guess a lot of folks (including me) are bumping into this issue. Including [ember inspector](https://github.com/Addepar/ember-table/issues/899#issuecomment-897195957).

Screenshot 2021-08-18 at 20 07 06

mixonic:
@MichalBryxi I'll bump the minor after https://github.com/Addepar/ember-table/pull/901 lands.

mixonic:
3.0.2 is out 🎉

srsgores:
For those who encounter this problem, our team encountered this very problem, and one of our team members discovered a fix in the form of using the value within `{{@row}}` as opposed to `{{@cell}}`.  Note that we **did** have to create a separate component for this purpose, thus using a new `get`er, though you could also opt to simply use `@row.cell.cellValue` directly in the template if you so choose (we merely did this for clarity).

srsgores:
We encountered this problem in the latest version (at the time of this writing, `4.0.0`)

PR #880, opened 2021-03-15
Pass rows count to sub-components

ahmacleod:
Passes the effective row count from each of `thead`, `tbody`, `tfoot` down through the component hierarchy. This count excludes rows that are hidden because their parent is collapsed, reflecting what the user actually sees. It is typically smaller than `rows.length`.

This can be used in the template to identify when the current row is the bottom rendered row.

Includes docs update and a supporting test.

Example usage:

```hbs
<EmberTable as |t|>
  <t.head @columns={{columns}} />
  <t.body @rows={{rows}} as |b|>
    <b.row as |r|>
      <r.cell as |cell column row cellMeta columnMeta rowMeta rowsCount|>
        {{#if (eq rowMeta.index 0)}}
          {{!-- custom first row markup --}}
          {{cell}}
        {{else if (lt rowMeta.index (dec rowsCount))}}
          {{!-- middle row markup; `dec` helper is from`ember-composable-helpers` --}}
          {{cell}}
        {{else}}
          {{!-- custom last row markup --}}
          {{cell}}
        {{/if}}
      </r.cell>
    </b.row>
  </t.body>
</EmberTable>
```

### DIFF
--- a/addon/components/-private/row-wrapper.js
+++ b/addon/components/-private/row-wrapper.js
@@ -88,6 +88,7 @@ export default Component.extend({
 
       let rowValue = this.get('rowValue');
       let rowMeta = this.get('rowMeta');
+      let rowsCount = this.get('rowsCount');
       let canSelect = this.get('canSelect');
       let checkboxSelectionMode = canSelect ? this.get('checkboxSelectionMode') : 'none';
       let rowSelectionMode = canSelect ? this.get('rowSelectionMode') : 'none';
@@ -116,6 +117,7 @@ export default Component.extend({
           rowMeta,
           rowSelectionMode,
           rowValue,
+          rowsCount,
         });
       });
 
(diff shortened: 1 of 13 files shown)


mixonic:
@ahmacleod can this PR also document the addition?

How do you access the index? Like, can you show an example of the usage?

ahmacleod:
> @ahmacleod can this PR also document the addition?
> 
> How do you access the index? Like, can you show an example of the usage?

Docs could definitely use a "how to tell which row this is from the template" section. This ties into https://github.com/Addepar/ember-table/pull/879.

Probably best to extend the "Table Meta Data" section, which currently does not even document the existing `rowMeta.index` property:

https://opensource.addepar.com/ember-table/docs/guides/main/table-meta-data

twokul (approved):

PR #881, opened 2021-03-16
Enforce min and max width constraints before initial fill

ahmacleod:
Ensures that min and max column widths are enforced on initial render.

Currently, min and max column widths are only enforced in two cases:
* when the `widthConstraint` mode is violated
* when the user manually resizes a column

In particular, if `widthConstraint` is `none`, or the chosen `widthConstraint` is not initially violated (e.g. `lte-container` when there is only one column), min and max column widths will be ignored on initial render.

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -676,6 +676,18 @@ export default EmberObject.extend({
       return;
     }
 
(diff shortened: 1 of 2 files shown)


jiayingxu (approved):
LGTM 🚀

PR #883, opened 2021-03-18
[WIP] Re-fill when scrollbar appears

ahmacleod:
WIP


ahmacleod:
Pulled this back into draft. A few changes I'd like to make:
* This should be an optional mode, not the default
* Needs docs
* Doesn't work in Firefox

PR #884, opened 2021-03-23
Replace hasBlock usage with has-block helper

kpfefferle:
Ember 3.26 introduces a deprecation for `hasBlock` in favor of `(has-block)`:

- RFC: https://github.com/emberjs/rfcs/pull/689
- PR: https://github.com/emberjs/ember.js/pull/19374
- Deprecation Guide: https://deprecations.emberjs.com/v3.x/#toc_has-block-and-has-block-params

 I was curious how easy this would be to address in ember-table, so this branch tries out the simple find-and-replace. It appears to pass all test scenarios *except* Ember 2.8. I'm not sure there's a polyfill for this (I'm not immediately finding one) to get us compatible back to 2.8, so maybe we just need to sit on it until we're ready to drop 2.8 compat - which I trust would happen before 4.0 drops the deprecated option?

### DIFF
--- a/addon/components/ember-tbody/template.hbs
+++ b/addon/components/ember-tbody/template.hbs
@@ -30,7 +30,7 @@
 
     as |api|
   }}
-    {{#if hasBlock}}
+    {{#if (has-block)}}
       {{yield (hash
         rowValue=api.rowValue
         rowMeta=api.rowMeta
--- a/addon/components/ember-td/template.hbs
+++ b/addon/components/ember-td/template.hbs
@@ -30,7 +30,7 @@
     {{/if}}
 
     <div class="et-cell-content">
-      {{#if hasBlock}}
+      {{#if (has-block)}}
         {{yield this.cellValue this.columnValue this.rowValue this.cellMeta this.columnMeta this.rowMeta this.rowsCount}}
       {{else}}
         {{this.cellValue}}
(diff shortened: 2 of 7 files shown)


twokul:
I think it is okay to hold off merging this PR until we drop Ember 2.8 as a target officially.

rwwagner90:
@kpfefferle @twokul We are getting a bunch of `hasBlock` deprecations in Ember Inspector, and I believe ember-table is the culprit. Any idea when this could be merged? Perhaps we could work on a `next` branch and support all the new things?

justinzelinsky:
> I think it is okay to hold off merging this PR until we drop Ember 2.8 as a target officially.

When do we think this will happen? I stumbled upon this when upgrading my app -- trying to get to Ember 3.26 and seeing warnings coming from latest `ember-table`.

rwwagner90:
@justinzelinsky you can use my fork, if you'd like https://github.com/rwwagner90/ember-table

kpfefferle:
This PR has been changed to target the `4.0-beta` branch since it requires dropping support for Ember 2.8

rwwagner90:
@kpfefferle since we are targeting the new branch, is there any reason not to go ahead and merge? Everything looks good to me!

mixonic:
:tada: great to see this get started!

PR #885, opened 2021-03-24
Update .gitignore

ahmacleod:
Adds `yarn.lock.ember-try` ~~and `.DS_Store`~~ to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,10 @@
 npm-debug.log*
 yarn-error.log
 testem.log
+.DS_Store
 
 # ember-try
 .node_modules.ember-try/
 bower.json.ember-try
 package.json.ember-try
+yarn.lock.ember-try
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -16,8 +16,10 @@
 npm-debug.log*
 yarn-error.log
 testem.log
+.DS_Store
```
kpfefferle, line 19: I literally just mentioned this in another PR earlier today (😆), but I'm personally of the opinion that since `.DS_Store` is a system-specific file and not a project-specific file, it belongs in a user's `~/.gitignore` and not in the project `.gitignore`

ahmacleod: I can be sold on this. @twokul @mixonic any strong feelings one way or the other?

mixonic (approved):
I think I do agree with @kpfefferle re: the macOS artifacts. Best done globally for your local user.

ahmacleod:
Settled. `.DS_Store` belongs in `~/.gitignore`.

PR #886, opened 2021-03-31
Specify meta key path for columns

ahmacleod:
Ember Table caches column meta data using the column object as a key. If a deeply equal, but not identical, column is substituted for the original, meta data such as the current column width (if not set explicitly on column) is lost. This makes it difficult to build the `columns` array in a purely functional style.

This PR allows the consumer to specify a `columnKeyPath` used to locate a key on the column object for caching meta data. For instance, if the `columnKeyPath` argument is set to `id`, Ember Table will use `column.id` as the key. This way, if Ember Table receives a new array of deeply equal, but not identical, columns, meta data will be preserved.

If `columnKeyPath` is used, we ~~throw an error~~ assert that no column has a missing/blank key, and that no two columns have the same key.

In practice, setting `columnKeyPath` to `valuePath` should be sufficient for most applications, assuming there are no duplicated columns. Again, this is an optional parameter; it is not required if your application is careful to retain its original column objects.

**TODO:**
- [x] Add unit tests for `MetaCache`
- [x] Assert that keys are unique within a given set of cols
- [x] Add failing test
- [x] Update docs


addon/components/ember-thead/component.js
```diff
@@ -281,15 +295,59 @@ export default Component.extend({
...
     if (this.get('columns.length') === 0) {
       return;
     }
+    this._validateUniqueColumnKeys();
     this._updateColumnTree();
 
     scheduleOnce('actions', this, this.fillupHandler);
   },
 
+  _validateUniqueColumnKeys() {
```
mixonic, line 313: Should this be written in an `assert` style so that it doesn't run in production?

mixonic (commented):
Oof.

addon/components/ember-thead/component.js
ahmacleod: I'm cool with that.

ahmacleod: Done.

jiayingxu (approved):

PR #901, opened 2021-08-11
Check for undefined attributeBindings

kpfefferle:
When we deploy our application using v3.0.2-3, we see the following console error:

```
TypeError: Cannot read property 'includes' of undefined
```

The stack trace points to https://github.com/Addepar/ember-table/blob/06ad94fc42cf68a87256faff494cf19dd84938c4/addon/components/ember-tbody/component.js#L273 which was introduced in https://github.com/Addepar/ember-table/pull/893

This PR adds an additional condition to check that `this.attributeBindings` is defined before calling `.includes()` on it. This fixes the error we see in our deployed application.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,8 @@ jobs:
             ember-lts-3.12,
             ember-lts-3.16,
             ember-lts-3.20,
+            ember-lts-3.24,
+            ember-production,
             ember-release,
             ember-beta,
             ember-canary,
--- a/addon/components/ember-tbody/component.js
+++ b/addon/components/ember-tbody/component.js
@@ -269,6 +269,11 @@ export default Component.extend({
     /*
      * Ember test selectors will remove data-test-row-count from the bindings,
      * so if it is missing there is no need to all the count.
+     *
+     * Even when ember-table is testing a production build, the test selectors
+     * addon remains enabled and causes `this.attributeBindings` to be present.
+     * In an actual app build, `this.attributeBindings` may be undefined.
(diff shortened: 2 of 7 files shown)


mixonic:
@kpfefferle is this for Ember 4.0? Can you ensure we have coverage of a reproduction?

kpfefferle:
@mixonic This happened in an app running Ember 3.27. This is the sort of thing that would likely be caught in a typed codebase as TS would notice that you are calling `.includes` on a property that could be `undefined`.

Do you have a suggestion how to test this case besides running our test suite against a production build (which then would drop the `data-test-*` selectors that many of our tests rely on?).

mixonic:
@kpfefferle we can use a production build of Ember but force the `data-test-` selectors to be enabled. We're actually doing something similar for internal Addepar codebases.

Let me take a stab at adding it to the build on this PR. I'd really like to make sure this has coverage.

shaun-addepar:
Hi, is there any update on this? We have a couple of blocker bugs([IX-322](https://addepar.atlassian.net/browse/IX-322) & [IX-323](https://addepar.atlassian.net/browse/IX-323)) in beta and staging that prevents all tables from rendering data and looks like it could be fixed by this PR.

mixonic:
I was unable to write a test for this as ember-test-selectors itself causes `attributeBindings` to be present even if using a production builds for test.

I added a comment explaining the patch and how it relates to tests, and I also did add a production build to the test suite here because why not.

mixonic:
Included in 3.0.2 🎉

Issue #897, opened 2021-07-23
Support Ember 4.0


mixonic:
Ember 4.0 is currently the canary channel, and it fails. There are likely several deprecations in 3.x which become removals in 4.0 and so break the library.

* [ ] 👇 Remove any deprecations which can be removed without breaking compat in `master`
* [ ] 👇 Identify deprecations which are a breaking change and consider them for a major version bump
* [x] `hasBlock` https://github.com/Addepar/ember-table/pull/884

kpfefferle:
One deprecation removal already in flight for `hasBlock`: #884

rwwagner90:
@mixonic I think ember-classy-page-object is one thing that is incompatible with modern Ember currently. I've been seeing babel errors for it. I do see the master branch has ember-cli-babel 7 though https://github.com/Addepar/ember-classy-page-object/blob/master/package.json#L23

Any plans to release a new version and update that here?

johanrd:
Supporting `Ember 4.0` may be a good occasion cut the cord for still supporting `Ember 2.8`.

In my view, the following could be a good fit for an `Ember Table 4.0` release:
• **Update dependencies** along the lines of the work being done at https://github.com/rwwagner90/ember-table 
• **Deprecate Ember global**, see https://github.com/rwwagner90/ember-table/pull/1
• **Replace hasBlock usage with has-block helper** (https://github.com/Addepar/ember-table/pull/884)
• **Deprecate Property Fallback Lookup** (no implicit this) (https://github.com/Addepar/ember-table/pull/907)
• **Use pre-built styles** (https://github.com/Addepar/ember-table/pull/900)

Regarding minimum ember-source support, updating to `"ember-test-selectors": 6.0.0` would for instance require at least `Ember 3.8` (as I understand it). 

There is seemingly enough momentum around this to not need to wait too long for a breaking release. As mentioned by @ctjhoa in https://github.com/Addepar/ember-table/pull/900 these can of course be portioned into several smaller breaking releases. However, I don't think we are very far from a breaking release that also supports `Ember 4.0`.

rwwagner90:
@johanrd the problem is, I think Addepar wants to keep the master branch and latest release in line with their supported versions for all of their apps. I have proposed that these addons are stable enough on their current versions, to support legacy Ember, but the problem is if new features want to be added, for example, something like column sorting, they want that to work across all their apps.

[...]

PR #907, opened 2021-09-09
Deprecate Property Fallback Lookup (no implicit this, take two)

johanrd:


mixonic:
I've rebased and force pushed this branch.

mixonic:
@johanrd This looks pretty great! The docs build is failing on what seems like a legit error (`.textContent` is being read off something which is now `null`) but the addon tests themselves are green 💪

johanrd:
@mixonic thanks, the `.textContent of null` error should now be resolved

kpfefferle (approved):
This was suggested in #897 as a candidate for 4.0, but it's a backwards-compatible change - so I'll go ahead and merge this to `master` and then rebase the `4.0-beta` branch to pick up on these changes since they do help unblock support for Ember 4.0.

PR #910, opened 2021-09-29
Add missing peer dependency @babel/core@^7.0.0-0

bantic:
When installing using `yarn`, there are many console warnings about an unmet peer dependency on `@babel/core@^7.0.0-0` for some of the deps, such as `@babel/plugin-proposal-object-rest-spread`.
These warnings also show up when consumers do their yarn install, e.g.:

```shell
warning "ember-table > @babel/plugin-proposal-nullish-coalescing-operator@7.14.5" has unmet peer dependency "@babel/core@^7.0.0-0".
```

This PR adds an explicit dependency on babel/core to silence those warnings.

Doing a `yarn install` in ember-table locally **before**:

image

And **after**:

image

I am taking inspiration from https://github.com/yarnpkg/yarn/issues/5347#issuecomment-463038189 that recommends that package maintainers add the missing peer dep locally, when the yarn warning originates from their package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "test:ember-compatibility": "ember try:each"
   },
   "dependencies": {
+    "@babel/core": "^7.0.0-0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "^7.12.13",
     "@babel/plugin-proposal-numeric-separator": "^7.12.13",
     "@babel/plugin-proposal-object-rest-spread": "^7.12.13",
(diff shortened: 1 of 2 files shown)


bantic:
Test failures appear to be due to changes in Ember v4 (which is "beta" at the time CI ran).

mixonic:
I've opened https://github.com/Addepar/ember-table/pull/911 to pin the CI at 3.28

PR #900, opened 2021-08-05
Use pre-built styles (#888)

ctjhoa:
This PR remove `ember-cli-sass` need as a direct dependency (still needed as devDependency) https://github.com/Addepar/ember-table/issues/888.
This is a quick win to solve the `ember-cli-sass` issue for an app that don't use it.
It is a straight conversion from scss to css using the official Dart Sass implementation.

How did I tests it:
```
$ git checkout master
$ yarn run build -- --output-path=dist-scss
$ git checkout remove_scss
$ yarn run build -- --output-path=dist-css
$ diff -qr dist-scss dist-css
Files dist-scss/assets/dummy.css.map and dist-css/assets/dummy.css.map differ
Files dist-scss/assets/vendor.css and dist-css/assets/vendor.css differ
$ diff dist-scss/assets/vendor.css dist-css/assets/vendor.css
129a130
>
134a136
>
139a142
>
144c147,149
< .ember-table td.is-fixed-left, .ember-table td.is-fixed-right,
---
>
> .ember-table td.is-fixed-left,
> .ember-table td.is-fixed-right,
150a156
>
154a161
>
158a166
>
162a171
>
165a175
>
168a179
>
171a183
>
175a188
>
179a193
>
182a197
>
189a205
>
192a209
>
199a217
>
204c222,224
< .ember-table .scroll-indicator__left, .ember-table .scroll-indicator__right {
---
>
> .ember-table .scroll-indicator__left,
> .ember-table .scroll-indicator__right {
208a229
>
211a233
>
216c238,240
< .ember-table .scroll-indicator__top, .ember-table .scroll-indicator__bottom {
---
>
> .ember-table .scroll-indicator__top,
> .ember-table .scroll-indicator__bottom {
220a245
>
224a250
>
228a255
>
231a259
>
238a267
>
241a271
>
247,248d276
<
< /*# sourceMappingURL=ember-table.css.map */
\ No newline at end of file
```
So basically, no changes.

`ember-cli-sass` as devDependency could be tackled later on.

PR #919, opened 2021-12-03
Ember Table 4.0 - with support for Ember 4.0

### LINKED ISSUE #897: Support Ember 4.0

kpfefferle:
The goal of this branch/PR is:

- [x] Drop Ember 2.8 & 2.12 support. The minimum supported version of Ember for ember-table 4.0 will be 2.18.
- [x] ember-cli-page-object will block supporting Ember 4 in the test helpers published with ember-table. There are a few options: Fix it upstream (there are some efforts), drop the dependency, drop the test helpers (this last option is unlikely to be the best as Addepar uses these extensively).
- [x] Re-enable the `ember-release` scenario which exercises Ember 4.0

In order to ship a release that supports Ember 4.0, we first will need to address any 3.x deprecations. https://github.com/Addepar/ember-table/issues/897 is a list to start with. PRs targeting this branch that move us in this direction are welcome.

Breaking change PRs for documenting:

- https://github.com/Addepar/ember-table/pull/884
- https://github.com/Addepar/ember-table/pull/900
- https://github.com/Addepar/ember-table/pull/921
- https://github.com/Addepar/ember-table/pull/940

The following upstream dependencies should be upgraded to stable releases when they occur:

- [ ] ember-classy-page-object 0.8
  - [ ] depends on ember-cli-page-object 2.0
- [ ] @html-next/vertical-collection 3.0

Resolves #897


kpfefferle:
I just published `4.0.0-beta.0` as an initial prerelease

MichalBryxi:
While trying the beta version:

```
    "ember-table": "4.0.0-beta.1",
```

I bumped into following error while running `ember server`:

```
❯ yarn start
yarn run v1.22.11
$ ember serve --proxy=http://localhost:3000

this.registerPlugin is not a function
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

I think I maybe might have fixed this by putting following to `package.json`:

```
  "resolutions": {
    "ember-cli-htmlbars": "6.0.1"
  },
```

MichalBryxi:
Another hurdle:

```
❯ yarn start
yarn run v1.22.11
$ ember serve --proxy=http://localhost:3000

Template Compiler Error (TemplateCompiler) in ember-table/components/-private/scroll-indicators/template.hbs

Cannot set property 'syntax' of undefined
```

Got "fixed" by putting following in `package.json`:

```
  "resolutions": {
    "ember-test-selectors": "6.0.0"
  },
```

kpfefferle:
@MichalBryxi Are you trying this with Ember 4? The errors you're seeing appear to be things that have not yet been addressed in this PR yet to work with Ember 4. The goal of this PR is to eventually get us there, but at this time even this beta branch is not yet compatible with Ember 4.

MichalBryxi:
@kpfefferle correct. I thought maybe this information might be helpful for someone trying to make it Ember-v4 compat.

package.json
```diff
@@ -66,6 +65,7 @@
     "ember-cli-deploy-git-ci": "^1.0.1",
     "ember-cli-eslint": "^5.1.0",
     "ember-cli-inject-live-reload": "^2.0.1",
+    "ember-cli-sass": "^10.0.0",
```
mixonic, line 68: Is this still needed?

kpfefferle: Good catch - that may have been inadvertently reintroduced during my rebase this morning

kpfefferle: @mixonic I believe that's still in use by the ember-table docs - which is why it's moved to `devDependencies`

kpfefferle: Everything in `tests/dummy/app/styles` is still written in Sass

balinterdi: In our fork, we replaced sass (scss) files with css and got rid of all the sass-related add-ons and scripts:
https://github.com/phorest/ember-table/commits/purge-sass

Would you consider accepting a PR to do that for this branch?

mixonic: Yeah I think so, if the change looks reasonable. It would only be a change to tests, as @kpfefferle mentioned the addon styles themselves have already been converted on this branch.

kpfefferle: This branch already removes the sass dependency from consuming apps which resolves the issues raised by users of the addon. I'm not sure we really have any compelling reason to remove it from the docs at this time 🤷‍♂️

mixonic: Seem ok to remove though, since the alternative is that we need to keep upgrading the dependency :-) Definitely not pressing.

balinterdi: Sorry, missed it that ember-cli-sass is removed in this PR, carry on 👏🏼

mixonic:
I've made fairly extensive updates to this branch this evening. The test suite and dependencies should be pretty much ready for Ember 4.0, although I'm not saying Ember 4.0 is green, or even that tests are totally unblocked.

* There probably remain tests using deprecated behaviors. These should be debugged and resolved on the 3.28 builds before wrestling with Ember 4.0 IMO.
* ember-cli-page-object will block supporting Ember 4 in the test helpers published with ember-table. There are a few options: Fix it upstream (there are some efforts on their repo), drop the dependency, drop the test helpers (this last option is unlikely to be the best as Addepar uses these extensively).
* There may be a complication around ember-test-selectors as well, I'm unsure.

Leaving it there for now. It should be a good foundation for someone else to dive into at this point.

kpfefferle:
@DanMonroe @mixonic Do any of the recently merged PRs need to be added to this PR's description list for specific call-out as breaking?

DanMonroe (approved):

kpfefferle:
`4.0.0-beta.4` has been released with full Ember 4 compatibility. We'll likely keep this in beta for a bit so people can try it out. Ideally, I'd hope to see the following dependencies release stable versions before we promote ember-table v4 to stable:

- [ ] ember-classy-page-object 0.8
  - [ ] depends on ember-cli-page-object 2.0
- [ ] @html-next/vertical-collection 3.0

vlascik:
How about forwarding `...attributes` to <table>, th, tr, tbody etc? Would make styling and integrating with frameworks like Bootstrap and Tailwind a lot easier.

mixonic:
Apologies to the vested parties for the hold-up. In our test suite at Addepar we noted some layout changes which were unexpected. It appeared that some columns were not keeping the same widths as in Ember Table 3, and row height might also be very, very slightly adjusted.

I was able to track this down today. In vertical-collection 2.0.0 (which our app relied upon) there was an errant `,` in `app.css`. This was resolved before `2.0.1`: https://github.com/html-next/vertical-collection/commit/fbb029c1078e84445f0a23150c11fdc43067dcd9

That extra comma caused a syntax issue in CSS, and effectively disabled those styling rules. This led the `display` of the occluded block to be `block`, and the `max-height` directive to have an impact (which it does not after the `display; table-row` from the disabled style block). That in turn impacts the sticky polyfill's calculation of "scale" for layout, corrupting columns and other sizes in very subtle and floating-point level ways.

Adopting the upgrade of ember-table bumped the vertical-collection version in our app, thus correcting the styles and re-enabling the `display: table-row` styles, and we got visual regression failures.

Having nailed that down, I believe we've addressed all concerns here. If there is any further feedback on getting 4.0 please chime in ASAP, if there is nothing new we will pull the trigger this week (hopefully tomorrow).

Thank you!

mixonic:
Ember 4.0 stable has been released! Thank you to to many contributors and stakeholders involved in this 🙇 

I would like to move forward pretty quickly with Ember Table 5.0. This would drop Ember 2.18, making the minimum supported Ember version 3.4-LTS. It would raise the minimum Node version. It would perhaps disable the "detect scaling styles" logic by default (as it corrupts layout logic) and make handling scaled UI opt-in (where it is currently a default behavior today). I'll open an issue soon, and I'm working on some of the scaling-related code changes.

PR #923, opened 2021-12-20
Spinner for infinite scroll

ahmacleod:
## Description

Adds an infinite scroll spinner component and a docs guide to implementing infinite scroll.

Details:
* Adds a `EmberTableLoadingMore` component that can be used to render a loading indicator beneath the table body
* Provides optional horizontal centering for the indicator that takes horizontal scroll position into account
* Adds an "Infinite Scroll" guide to the docs with recipe for implementing infinite scroll
* Does not actually provide a spinner .gif or element; consumer specifies one via block

## Demo

20211221 Ember Table Infinite Scroll

### Infinite Scroll Guide

image
image
image
image

### API docs

image
image


addon/components/ember-table-loading-more/component.js
```diff
@@ -0,0 +1,175 @@
...
+    this.canLoadMoreChanged();
+    this.isLoadingChanged();
+    this.centerChanged();
+  },
+
+  willDestroyElement() {
+    if (this.get('center')) {
+      this.removeListeners();
+    }
+    this._super(...arguments);
```
twokul, line 111: Is there a reason to call `super` last?

twokul (approved):
The code looks good. Left one small comment!

✅

addon/components/ember-table-loading-more/component.js
ahmacleod: It's an old-school LIFO/OO type thing. When initializing, we want to call `super` first to allow the parent class to do its setup before the child. When tearing down, we do the reverse. In this case it doesn't make any difference, but generally speaking, the parent might tear down some things that are needed for the child to do its (more specific) cleanup.

PR #925, opened 2022-01-12
Basic dependencies and CI for Ember 4 compat

mixonic:
* If working on this, be careful about addressing deprecations in the test harness. You don't want to adapt APIs unavailable in the older library versions required by Ember 2.18
* In this PR as it stands I've regressed the CI suite going green on 3.28. A more incremental approach (like the many PRs at https://github.com/html-next/vertical-collection/pulls?q=is%3Apr+sort%3Aupdated-desc+is%3Amerged) is likely wise.
* ember-cli-page-object will block supporting Ember 4 in the test helpers published with ember-table. There are a few options: Fix it upstream (there are some efforts), drop the dependency, drop the test helpers (this last option is unlikely to be the best as Addepar uses these extensively).


mixonic:
This PR is supplanted by the three PRs referencing it.

PR #927, opened 2022-01-14
Bump dependencies, configure Ember for 4.0

mixonic:
Extracted from https://github.com/Addepar/ember-table/pull/925

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
   test:
     name: Run Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 6
+    timeout-minutes: 7
 
     steps:
       - name: Checkout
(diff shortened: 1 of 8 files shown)


config/ember-try.js
```diff
@@ -13,6 +13,9 @@ module.exports = function() {
         {
           name: 'ember-lts-2.18',
           npm: {
+            dependencies: {
+              'ember-test-selectors': '5.0.0',
+            },
```
kpfefferle, line 18: Is this needed if the default in `package.json` is `^5.0.0`, or does locking 2.18 to *exactly* `5.0.0` make a difference?

kpfefferle (approved):

config/ember-try.js
mixonic: It is needed exactly. For our 2.12 app at Addepar we use resolutions to pin it. I could use resolutions here as well, maybe.

Going to land this and move on though, I have a few more unblocking-other-people changes I think I can get in here quickly.

mixonic: nm, need to rebuild anyway so trying to drop it. Will try resolutions if it is needed.

mixonic: Well, seems like I was wrong. This must only be a limitation of 2.12! :tada:

PR #929, opened 2022-01-14
Update test helpers for Ember 4.0

mixonic:
Extracted from https://github.com/Addepar/ember-table/pull/925 and refined.

* Note that this change **removes** test coverage of the production build output that was added in https://github.com/Addepar/ember-table/pull/901. Upon reviewing that PR that test run isn't critical.


tests/unit/-private/table-sticky-polyfill-test.js
```diff
@@ -4,13 +4,13 @@ import { A as emberA } from '@ember/array';
...
-import { find, findAll, scrollTo } from 'ember-native-dom-helpers';
-import wait from 'ember-test-helpers/wait';
+import scrollTo from '../../helpers/scroll-to';
+import { find, findAll, settled } from '@ember/test-helpers';
 
 import { setupTableStickyPolyfill } from 'ember-table/-private/sticky/table-sticky-polyfill';
 
-const HEADER_PIXEL_EPSILON = 10;
-const FOOTER_PIXEL_EPSILON = 10;
+const HEADER_PIXEL_EPSILON = 15;
```
mixonic, line 12: Here is the weakened assertion.

mixonic: Well, this didn't result in a successful run. Going back to `10`.

PR #940, opened 2022-02-15
Add ember-try scenario for ember v4

DanMonroe:
This PR requires an update to ember-class-page-object.  It is pointing to a specific beta branch based that points to a beta branch of https://github.com/san650/ember-cli-page-object/tree/v2.0.0-beta.3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,8 @@ jobs:
 
     strategy:
       matrix:
-        ember-version: [
+        ember-version:
+          [
             ember-lts-2.18,
             ember-lts-3.4,
             ember-lts-3.8,
@@ -51,10 +52,9 @@ jobs:
             ember-lts-3.20,
             ember-lts-3.24,
             ember-lts-3.28,
-            ember-4.0,
-            # ember-release,
+            ember-release,
             # ember-production,
-            ember-default-docs,
+            ember-default-docs
           ]
 
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,8 +42,7 @@ jobs:
 
     strategy:
       matrix:
-        ember-version:
-          [
+        ember-version: [
             ember-lts-2.18,
             ember-lts-3.4,
             ember-lts-3.8,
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -84,6 +84,10 @@ rows: computed(function() {
 });
 ```
 
+### Using ember-table with Ember <= 3.24
+Use resolutions in your package.json to pin down `ember-classy-page-object` to version `0.7.0`.
+Newer versions are used to support Ember >= 3.28
+
```
kpfefferle, line 90: I'd suggest we move this up higher in the README to be included with the installation instructions.

config/ember-try.js
```diff
@@ -80,6 +101,14 @@ module.exports = function() {
...
           },
         },
+        {
+          name: 'ember-4.0',
+          npm: {
+            devDependencies: {
+              'ember-source': '~4.0.1',
+            },
+          },
+        },
```
kpfefferle, line 111: Can we just re-enable the `ember-release` scenario instead of adding a `4.0.x`-specific scenario? That could keep us on the path of only carving out specific scenarios for supported LTS versions.

tests/unit/-private/column-tree-test.js
```diff
@@ -50,7 +33,6 @@ module('Unit | Private | ColumnTree', function(hooks) {
     tree = ColumnTree.create({
       columns,
       columnMetaCache,
-      rows: generateTree([0, [1, [2, 3], 4, [5, 6]]]),
```
kpfefferle, line 53: This was causing some sort of error, but seems irrelevant to the test, right? A bit of explanation in a comment would be helpful for review.

.github/workflows/ci.yml
```diff
@@ -52,9 +51,10 @@ jobs:
             ember-lts-3.20,
             ember-lts-3.24,
             ember-lts-3.28,
+            ember-4.0,
```
kpfefferle, line 54: I'd like to see us re-enable `ember-release` here

.github/workflows/ci.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
     strategy:
       matrix:
-        ember-version:
+        ember-version:
```
kpfefferle, line 45: Somehow added a trailing space to this unchanged line?

```suggestion
        ember-version:
```

kpfefferle (approved):

PR #943, opened 2022-02-23
Use ember-classy-page-object v0.8.0-beta

DanMonroe:
Uses a beta release of ember-classy-page-object that uses ember-cli-page-object v2.0.0-beta.3


kpfefferle (commented):
I thought that this update was intertwined as part of https://github.com/Addepar/ember-table/pull/940? As long as we can land it as part of that PR, I don't think we need this one.

PR #948, opened 2022-05-03
Refactor offsetHeight use float-compat API

mixonic:
`offsetHeight`, per MDN, is an API which returns _integer_ values. This makes is inappropriate to use in calculations of current UI scale (usually from a transform) as it rounds any floating point height into an integer.

This PR changes the `getScale` function to use `getComputedStyle` to read off the logical height of the element. This API supports greater precision (to at least 3 digits) though it also may not provide the same level of precision as `getBoundingClientRect`.

This prevents the positioning data for cells (which is impacted by `scale`) from being corrupted by heights which are at decimal precision.

### DIFF
--- a/addon-test-support/pages/-private/ember-table-header.js
+++ b/addon-test-support/pages/-private/ember-table-header.js
@@ -139,7 +139,7 @@ const Header = PageObject.extend({
     /**
      * Below a certain number of steps, Hammer (the gesture library
      * which recognizes panning) will not pick up the pointermove
-     * events emitted by `mouseMove` before the gestrure completes.
+     * events emitted by `mouseMove` before the gesture completes.
      *
      * 5 seems a reasonable number.
      */
--- a/addon/-private/utils/element.js
+++ b/addon/-private/utils/element.js
@@ -30,18 +30,72 @@ export function closest(el, selector) {
   return null;
 }
 
+/*
+ * Calculate the scale of difference between an element's logical height
+ * (the offsetHeight or getComputedStyle height) compared to its rendered
+ * height (via getBoundingClientRect).
+ *
+ * Note that there are some interesting edge cases to consider around
+ * these APIs.
+ *
+ *   - `getComputedStyle` returns a string from styles in pixels. For
+ *     example `48.23px`. The precision of this API in Chrome seems to
+ *     be 3 decimal places. Additionally, there can be unexpected values
+ *     such as `auto`. Finally, this API may be slow compared to other
+ *     DOM API calls.
+ *   - `offsetHeight` always returns in integer. If the height of an
+ *     element is at float precision, then the value is rounded. This
+ *     makes it at best useful as a fallback.
+ *   - `getComputedStyle` may return a smaller value than expected when
+ *     used with exotic kinds of styling (likely due to passing/broder
+ *     and block model or overflow issues?). In the ember table codebase
+ *     today, only the `table` tag itself is passed in for measurement.
+ *
+ */
 export function getScale(element) {
   let rect = element.getBoundingClientRect();
+  let renderedHeight = rect.height;
 
-  if (element.offsetHeight === rect.height || rect.height === 0) {
+  if (renderedHeight === 0) {
+    return 1;
+  }
+
+  let computedHeightStyleValue = window.getComputedStyle(element).height;
+  let computedHeightInPixels = Number(
+    computedHeightStyleValue.substring(0, computedHeightStyleValue.length - 2)
+  );
+
+  let offsetHeight = element.offsetHeight;
+
+  if (isNaN(computedHeightInPixels)) {
+    computedHeightInPixels = offsetHeight;
+  } else if (Math.abs(computedHeightInPixels - offsetHeight) >= 1) {
+    throw new Error(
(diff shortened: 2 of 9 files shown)


addon/-private/utils/element.js
```diff
@@ -30,13 +30,45 @@ export function closest(el, selector) {
...
+  let renderedHeight = rect.height;
 
-  if (element.offsetHeight === rect.height || rect.height === 0) {
+  if (renderedHeight === 0) {
+    return 1;
+  }
+
+  let computedHeightStyleValue = window.getComputedStyle(element).height;
+  let computedHeightInPixels = Number(
+    computedHeightStyleValue.substring(0, computedHeightStyleValue.length - 2)
```
c69-addepar, line 61: can we use `parseFloat` here, maybe ?

also, how would this handle the `auto`, `inherit`, `initial` and similar non-pixel string values ?

addon/-private/utils/element.js
```diff
@@ -30,13 +30,45 @@ export function closest(el, selector) {
...
+  );
+
+  if (isNaN(computedHeightInPixels)) {
```
c69-addepar, line 64: oh, my bad ))) 
but then - parseFloat makes even more sense.

And maybe we can explain in a comment - why NaN is a possible value here.

tests/unit/-private/element-utils-test.js
```diff
@@ -0,0 +1,33 @@
...
+
+    div.style.transform = 'scale(0.5)';
+
+    assert.equal(getScale(div), 2, 'scale on a scaled element is correct');
+
+    div.style.height = '1.5px';
+
+    assert.equal(getScale(div), 2, 'scale on a scaled element is correct');
+  });
+});
```
c69-addepar, line 33: what about the "normal scenario" test ?

addon/-private/utils/element.js
mixonic: `parseFloat` docs suggest `Number(string)` as having stricter parsing: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat I don't think there is a reason to use `parseFloat` over `Number`.

`isNaN(Number('a')) === true`, hence the `isNaN` check on the next line. I can leave a comment about this.

tests/unit/-private/element-utils-test.js
mixonic: This test contains three assertions:

* scale of an element with no transforms is `1`.
* scale of an element with integer height by `0.5` is `2`
* scale of an element with float height by `0.5` is `2`

The final case is what fails without the rest of this patch.

addon-test-support/pages/-private/ember-table-header.js
```diff
@@ -29,18 +40,52 @@ const Header = PageObject.extend({
 
   /**
    * Retrieves selected header cell width.
+   *
+   * Deprecated: offsetWidth returns a rounded integer, and so can
```
c69-addepar, line 44: nit: use `@deprecated` jsdoc tag

addon-test-support/pages/-private/ember-table-header.js
```diff
@@ -29,18 +40,52 @@ const Header = PageObject.extend({
...
+   * result in unreliable tests. Use logicalWidth.
    */
   get width() {
     return findElement(this).offsetWidth;
   },
 
   /**
    * Retrieves selected header cell height.
+   *
+   * Deprecated: offsetHeight returns a rounded integer, and so can
```
c69-addepar, line 54: nit: same

addon-test-support/pages/-private/ember-table-header.js
```diff
@@ -58,10 +103,24 @@ const Header = PageObject.extend({
 
   contextMenu: triggerable('contextmenu'),
 
-  /**
-   * Resizes this column by dragging right border several pixels.
+  /*
+   * This API is deprecated, prefer the more explicit resizeLogicalSize.
```
c69-addepar, line 107: same

addon/-private/utils/element.js
```diff
@@ -30,13 +30,45 @@ export function closest(el, selector) {
...
 export function getScale(element) {
   let rect = element.getBoundingClientRect();
+  let renderedHeight = rect.height;
 
-  if (element.offsetHeight === rect.height || rect.height === 0) {
+  if (renderedHeight === 0) {
+    return 1;
+  }
+
+  let computedHeightStyleValue = window.getComputedStyle(element).height;
```
c69-addepar, line 59: do we have a performance test for rendering a large table (100 visible rows) with dynamic height ? Or maybe you have done it manually, if automated cannot catch slowdowns ?

because i am pretty sure you are aware of the dangers (killing performance) with `getComputedStyle`, but will still put an article (not THE article, just AN article) link here:

https://pagebuildersandwich.com/increased-plugins-performance-200/

mmun (approved):

addon/-private/utils/element.js
mixonic: I've addressed this. `getScale` is now only called once and is passed through the system where needed.

mixonic:
I've confirmed in an internal codebase that this improves how well a table can match the intended container it is in. Landing it, and I'll but a beta/preview release for Ember Table v6 sometime soon.

mixonic:
Released in v6.0.0-0 prerelease

Issue #578, opened 2018-07-12
Add ability to apply classes to table element?


bgentry:
Some CSS frameworks (i.e. [Bootstrap](https://getbootstrap.com/docs/4.0/content/tables/)) work by applying classes to the `table` element:

```html
<table class="table table-sm">
```

However, ember-table does not currently offer a way to customize the `table` element in any way, [rendering only a static `<table>` tag](https://github.com/Addepar/ember-table/blob/0b77cdf1ac2ba50166202083f5182a4870294a8c/addon/components/ember-table/template.hbs).

Can we add the ability to add classes to the `table` itself? Passing `<EmberTable class="table">` actually just applies the classes to the wrapping `<div>` rather than the `<table>`. Maybe an optional argument to the component called `tableClass`?

```hbs
<EmberTable @tableClass="table table-sm">
```

bgentry:
I realized right after posting this that the Bootstrap styles actually do still work when the classes applied to the wrapping `<div>` 🤦‍♂️ . Gonna go ahead and close this, feel free to reopen if you do want this feature.

pzuraq:
Glad it worked out for you. Just wanted to address this real quick in case people are wondering in the future, I'm generally against passing classes in as arguments (other than outerHTML semantics with `class`). It's not sustainable as an API practice, because it leads to an explosion of API space (e.g. adding an argument for every conceivable element that could be customized).

This is problematic for theming and especially for using CSS frameworks that come with styles for a specific HTML structure OOTB. I think this is a much more general problem though, and we don't really have great solutions for it. Currently, my favored path is using CSS preprocessors like SASS to fit the styles to the HTML structure. Well designed SASS/LESS/Stylus frameworks should make this doable, but not all off-the-shelf style frameworks are well designed 😕 

Maybe the answer is we do provide class name arguments for some things that are very likely to be customized (`table` tag, etc), but I'd like to explore other options before we commit to that as an API paradigm.

jelhan:
> I realized right after posting this that the Bootstrap styles actually do still work when the classes applied to the wrapping

Only some of the styles are working for Bootstrap 4. E.g. the table is not using full width if `.table` is only present on wrapping `<div>`. It's not working at all for Bootstrap 3.

@pzuraq Please consider to reopen this one. I have to override template for `<EmberTable>` to apply a class required by Bootstrap. :disappointed:

cluis13915:
[...]

PR #953, opened 2022-06-15
Add custom class names to table element

kevintraver:
This fixes https://github.com/Addepar/ember-table/issues/578

```
<EmberTable @tableClassNames="my-custom-class" as |t|>
  <t.head @columns={{this.columns}} />

  <t.body @rows={{this.rows}} />
</EmberTable>
```

### DIFF
--- a/addon/components/ember-table/template.hbs
+++ b/addon/components/ember-table/template.hbs
@@ -3,7 +3,7 @@
   class="ember-table-overflow"
   id="{{this.elementId}}-overflow"
 >
-  <table>
+  <table class="{{this.tableClassNames}}"
     {{yield (hash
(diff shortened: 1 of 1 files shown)


mixonic:
@kevintraver I don't think I can accept this change as put forward here. I'd prefer to see `...attributes` used and then `<EmberTable class="my-custom-class" />` becomes possible. Generally that direction, and furthermore a direction of using glimmer or template only components to decouple from the DOM entirely, are the direction this work should head.

I'm going to close this to express that this specific API isn't one the project is interested in, but I appreciate your contribution and would be happy to chat over next steps sometime if you want to get into it!

Issue #957, opened 2022-07-21
Proposal: Ember Table 5


mixonic:
Building atop https://github.com/html-next/vertical-collection/issues/362 and https://github.com/html-next/vertical-collection/pull/372, I propose a 5.0 release of Ember Table.

* Support Ember 3.12+
* Require vertical-collection 4.x
* Change various calculations which rely on browser APIs that were presumed to return accurate values, but actually return rounded integers: https://github.com/Addepar/ember-table/pull/948
* Register the raf waiter as part of the setupForTest API https://github.com/Addepar/ember-table/issues/956
* Only document/test `<EmberTable />` style invocation over curly invocation.

What else?

acorncom:
Those things look good on this end 👍

Issue #956, opened 2022-07-11
Add ember-raf-scheduler registerWaiter using `setupForTest` or document it?


acorncom:
We've recently begun using ember-table in several client apps (and quite liking it, thanks for all the hard work!). One thing we've started noticing after we introduced it was that we're seeing flakey tests. It's been hard to get to the bottom of what was going on, but when seeing this across three different apps of varying ages and on various versions of newer Ember, it didn't seem like it was app-specific.

I dug in further today and it looks like there have been intermittent discussions of the same issue by a number of different folks (see https://github.com/Addepar/ember-table/issues/810, https://github.com/Addepar/ember-table/issues/828, https://github.com/Addepar/ember-table/issues/667). Suggestions were to pass `showAll=true` to `vertical-collection` to disable occlusion (which was where we were seeing snags) but that seemed a bit like an ugly hack. Finally found https://github.com/html-next/vertical-collection/issues/231 which referenced registering a test waiter for `ember-raf-scheduler` and when that was added in in our app all "flakiness" went away.

Should we consider adding the `ember-raf-scheduler#registerWaiter` call as part of the `setupForTest` that `ember-table` publishes or note the potential need for it for testing purposes in the test section of the readme? I'm happy to do the work, but thought I'd ask what your preferences would be for the library

PR #963, opened 2022-08-29
Ember Table 5

twokul:
Fixes https://github.com/Addepar/ember-table/issues/957.

A list of changes:

- [x] Remove Ember version < 3.12
- [x] Bump dependencies
- [x] Update supported version in README
- [x] Remove the "removal of contextual components"
- [x] Default to `_notifyPropertyChange`
- [x] Revert to mainstream testing setup
- [x] Use the `scrollTo` helper from `@ember/test-helpers`
- [x] Unwrap runloop
- [x] Bump the version to 5-beta
- [x] Bump ember-classy-page-object
- [x] Angle-bracket syntax for components
- [x] Bump `vertical-collection` to v4
- [x] Address https://github.com/Addepar/ember-table/issues/956


addon-test-support/index.js
```diff
@@ -1,10 +1,11 @@
 import TablePage from './pages/ember-table';
-
+import registerWaiter from 'ember-raf-scheduler/test-support/register-waiter';
```
kpfefferle, line 2: 👍

addon/components/ember-tbody/template.hbs
```diff
@@ -1,4 +1,5 @@
-{{#vertical-collection this.wrappedRows
+{{#vertical-collection
```
kpfefferle, line 1: Any reason not to convert our usage of vertical-collection to angle brackets while we're at it?

config/ember-try.js
```diff
@@ -10,62 +10,17 @@ module.exports = function() {
...
-        },
         {
           name: 'ember-lts-3.12',
           npm: {
             devDependencies: {
               'ember-source': '~3.12.0',
             },
             dependencies: {
-              'ember-classy-page-object': '0.7.0',
+              'ember-classy-page-object': '0.8.0-beta.2',
```
kpfefferle, line 20: If these all match the default version found in `package.json` now, then we shouldn't need to override them here, right?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-table",
-  "version": "4.0.0",
+  "version": "5.0.0-beta.0",
```
kpfefferle, line 3: Can we publish this beta to NPM so people can easily try it with their apps? Or are we planning to cut the 5.0 stable version soon?

kpfefferle (approved):
Minor comments, but overall LGTM

package.json
mixonic: Yeah, I don't see this on public NPM.

addon/components/ember-tbody/template.hbs
mixonic: Vertical collection is mostly using and tested with `<VerticalCollection` upstream 👍 The only reason not to use that style would be if a block has an `{{else}}`.

config/ember-try.js
mixonic: https://github.com/Addepar/ember-table#using-ember-table-with-ember--324 is probably still relevant, though obviously the test suite is green. Can you rationalize this @twokul? I think we were limited to `0.7.0` on Ember < 3.24 based on upstream support in ember-cli-page-object.

addon/components/ember-tbody/template.hbs
twokul: We cannot quite do it yet because the `{{else}}` block is not supported in angle bracket syntax (and does not have the same equivalent.)

config/ember-try.js
twokul: Yeah, you are 100% correct. I think we can just "remove and forget." Not sure how I missed this.

package.json
twokul: I have not published anything yet. Mostly, just wanted a review + sanity check on the changes :)

twokul:
Published [`v5.0.0-beta.0`](https://www.npmjs.com/package/ember-table/v/5.0.0-beta.0)

twokul:
Published [`v5.0.0-beta.1`](https://www.npmjs.com/package/ember-table/v/5.0.0-beta.1)

twokul:
Published [`v5.0.0-beta.2`](https://www.npmjs.com/package/ember-table/v/5.0.0-beta.2)

PR #975, opened 2023-01-26
Use component template colocation instead of pods

rwwagner90:
I'm not sure if there is a reason these need to still be pods or not, but refactoring to colocation fixes some issues with embroider and might be a good step forward for this addon.


mixonic:
Template co-location is definitely a good incremental step forward on the path to full v2 addonization!

PR #976, opened 2023-02-07
Only remove ResizeSensor if it exists

apellerano-pw:
Repro:
```hbs
<t.loadingMore
  @isLoading={{true}}
  @center={{false}}
/>
```

1. The `didInsertElement` hook fires,
2. it calls `this.centerChanged`,
3. which calls `this.removeListeners`, since `center` is false.

The problem is that `addListeners` has never been called, and `_scrollElementResizeSensor` doesn't exist. So you get a JS console error.

One possible fix is in this PR.

There's technically a workaround, but it's stomach-turning:

```hbs
<t.loadingMore
  @isLoading={{true}}
  @center={{this.eventuallyFalse}}
  {{did-insert (action (mut this.eventuallyFalse) false)}}
/>
```

Let `@center` start as true, and then set it to false slightly later... now `addListeners` was called, so there's no console error and you eventually get the state you want 😬

### DIFF
--- a/addon/components/ember-table-loading-more/component.js
+++ b/addon/components/ember-table-loading-more/component.js
@@ -144,7 +144,9 @@ export default Component.extend({
 
   removeListeners() {
     this.get('scrollElement').removeEventListener('scroll', this._updateTransform);
-    this._scrollElementResizeSensor.detach();
+    if (this._scrollElementResizeSensor) {
+      this._scrollElementResizeSensor.detach();
+    }
   },
 
   updateTransform() {


apellerano-pw:
As an alternative, the whole display aspect of `ember-table-loading-more` may be a good cut candidate. The implementation is heavy on resize performance, with the `ResizeSensor` usage. And end users can implement that functionality easily themselves with `position: absolute` or the like

kpfefferle (approved):
Good catch. This is the sort of thing I love TypeScript for since it would have made this pretty apparent right away 🙂

kpfefferle:
Rebased onto `master` which removed the 3.16 CI scenario that kept timing out ✅

kpfefferle:
@apellerano-pw Released in v5.0.2 - thanks! 🚀

PR #979, opened 2023-04-06
Documentation site upgrade

mixonic:
* Upgrade ember-cli-addon-docs
  * This addresses the ugly, unreadable red background on many of the docs pages.
* Move ember-cli-deploy to only be a dependency when doing a deployment
* Migrate `pods` docs app layout to traditional Ember app layout
* Bump Node and Yarn for Volta
* Remove many hacks, some of which may need to come back once I look at the test suite.


c69-addepar (approved):
LGTM

PR #980, opened 2023-04-10
Restore SPA implementation

mixonic:
Based on https://github.com/rafgraph/spa-github-pages, update the index and 404 pages to allow history-like URLs to work properly.


bantic (approved):
LGTM

PR #985, opened 2023-04-24
Fix keyUp event bug in ember-th

chiaoyu-addepar:
In `addon/components/ember-th/component.js` there is a keyUp event which will trigger `updateSort` function with Enter key. However, the current code will show the error in the below screen shot when user hitting Enter ( when focusing on the header columns )
This PR is trying to fix this bug with providing a `toggle: false` in `this.updateSort` when being called in the keyUp event.

Screenshot 2023-04-24 at 8 51 29 AM

### Reviewer note
not sure if we want to support the `toggle` the same way as in the click event
`let toggle = event.ctrlKey || event.metaKey;`


mixonic:
@chiaoyu-addepar the build is unblocked by https://github.com/Addepar/ember-table/pull/986. Please give this a rebase.

c69-addepar (approved):
LGTM

PR #987, opened 2023-04-28
Make Ember Table header accessible through tab and maintain focus state after trigger sorting

chiaoyu-addepar:
In this PR, added a tabindex to the attributeBinding in order to make Ember Table header be able to navigate/get focus  with `tab` key.
After the using `Enter` key to trigger the sorting, in currently code base, the table header will loose focus since all the component (tr, th) will re-render.
In this PR, added a `key=` field in {{#each}} to solve the focus maintain issue. 
With this solution, `didInsertElement` function in `th` component will not be called each time after sorting, which will cause problem for reordering as it do `this.get('columnMeta').registerElement(this.element);` in the `didInsertElement` function which will be used when doing reorder.
To fix this, added a `didRender` function and do `this.get('columnMeta').registerElement(this.element);` there.

### DIFF
--- a/addon/components/ember-th/component.js
+++ b/addon/components/ember-th/component.js
@@ -103,8 +103,6 @@ export default BaseTableCell.extend({
   didInsertElement() {
     this._super(...arguments);
 
-    this.get('columnMeta').registerElement(this.element);
-
     let hammer = new Hammer(this.element);
 
     hammer.add(new Hammer.Press({ time: 0 }));
--- a/addon/components/ember-th/component.js
+++ b/addon/components/ember-th/component.js
@@ -37,7 +37,7 @@ const COLUMN_REORDERING = 2;
 export default BaseTableCell.extend({
   layout,
   tagName: 'th',
-  attributeBindings: ['columnSpan:colspan', 'rowSpan:rowspan'],
+  attributeBindings: ['columnSpan:colspan', 'rowSpan:rowspan', 'tabindex:tabindex'],
   classNameBindings: ['isSortable', 'isResizable', 'isReorderable'],
 
   /**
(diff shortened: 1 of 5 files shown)


addon/components/ember-th/component.js
```diff
@@ -115,6 +117,10 @@ export default BaseTableCell.extend({
     this._hammer = hammer;
   },
 
+  didRender() {
+    this.get('columnMeta').registerElement(this.element);
+  },
```
kpfefferle, line 122: Upon initial render, [`didRender` gets called immediately following `didInsertElement`](https://guides.emberjs.com/v3.12.0/components/the-component-lifecycle/). If we're going to execute this line of code in `didRender`, then I don't believe that we should *also* be calling it above in `didInsertElement`.

chiaoyu-addepar: Thanks for catching this.

addon/components/ember-th/component.js
```diff
@@ -61,6 +61,8 @@ export default BaseTableCell.extend({
 
   rowMeta: readOnly('api.rowMeta'),
 
+  tabindex: 0,
```
mixonic, line 64: If this value is only ever `0`, then in `didInsertElement` can you use `setAttribute(` to set it instead of a binding? A binding has setup/memory overhead, a static attribute setup would be better for performance (and not really more confusing or anything).

addon/components/ember-th/component.js
mixonic: Why is this being moved to `didRender`?

addon/components/ember-tr/template.hbs
```diff
@@ -1,4 +1,4 @@
-{{#each this.cells as |api|}}
+{{#each this.cells key="@index" as |api|}}
```
mixonic, line 1: this changes the stability test not just for headers, but for all rows.

mixonic (commented):
Can this be tested?

PR #990, opened 2023-05-17
Disabled "ember-release" scenario in ember-try

c69-addepar:
This PR disabled "ember-release" scenario in ember-try, as it started to return `5.0.0` and fail the CI for all new PRs.

Proper `5.0.0` support will be fixed and this check will be re-introduced in another PR for same ticket.

Ticket: WEBCORE-2069

Release Notes: We have discovered that ember-table is currently incompatible with ember 5.0.0 due to breaking changes in upstream

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,8 @@ jobs:
             ember-lts-3.24,
             ember-lts-3.28,
             ember-lts-4.4,
-            ember-release,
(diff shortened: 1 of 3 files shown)


c69-addepar:
> That support can be fixed and re-introduced in another PR though.

That's the plan!

DanMonroe (approved):
LGTM

PR #991, opened 2023-05-18
Refactor helper function function to test the semi-automatic refactoring

c69-addepar:


c69-addepar:
Hmm.. IDE does not detect lint errors..

image

PR #992, opened 2023-06-29
fix: import htmlSafe from @ember/template, not @ember/string

ascudder:
This fixes the deprecation in Ember 3.25 https://deprecations.emberjs.com/v3.x/#toc_ember-string-htmlsafe-ishtmlsafe

Currently this is blocking our ability to use this addon in an Ember 3.28 application.

### DIFF
--- a/addon/components/ember-table/component.js
+++ b/addon/components/ember-table/component.js
@@ -1,6 +1,6 @@
 import Component from '@ember/component';
 import { computed } from '@ember/object';
-import { htmlSafe } from '@ember/string';
+import { htmlSafe } from '@ember/template';
 import {
   setupTableStickyPolyfill,
   teardownTableStickyPolyfill,


kpfefferle:
> Currently this is blocking our ability to use this addon in an Ember 3.28 application.

I'm not sure why this would block usage in an Ember 3.28 application given that deprecations aren't removed until the next major release (4.0) 🤔

I'm also not sure why I never saw this issue in our internal application that runs on 4.x and makes extensive use of ember-table. I do remember it being an issue with some of our other addons when we were trying to address deprecations in 3.28 before upgrading to 4.x.

kpfefferle:
Note also that our CI tests on 3.28, 4.4, and 4.12 and this is does not cause issues with those test sceanarios

ascudder:
@kpfefferle The only difference I could think of off the top of my head is that it's a test running in an addon that we've written that uses ember-table. That addon is also currently on 3.28 as well. 

By patching the component locally, I was able to fix the errors in the tests.

Since version 5.0 of Ember Table says it's compatible with versions of Ember after 3.8 (which is when this deprecation was added), is there any reason not to patch this? Can we please get this merged to fix the issue?

kpfefferle:
@ascudder I'm going to do a couple sanity checks with some of our older and larger internal apps just to make sure this doesn't cause us issues. Barring any surprises found there, I don't have any issue merging this 👍

kpfefferle (approved):
Confirmed this doesn't cause any issues in our internal apps 👍

kpfefferle:
@ascudder Merged and released as `v5.0.5` 🚀

ascudder:
Thank you @kpfefferle ! This is a big help for us as we work toward upgrading as well.

PR #994, opened 2023-07-30
Fix a memory leak with the root column tree node

mmun:
This PR fixes a memory leak that due to the root node not being destroyed when its dependency key `columns` is invalidated.

The reason this leaks memory is because some nodes in the ColumnTree structure add observers, and those observers are stored in a global map. If we don't explicitly destroy them the observers are never removed.

This exact pattern is already in use in this codebase for the CollapseTree: https://github.com/Addepar/ember-table/blob/master/addon/-private/collapse-tree.js#L810-L812

### DIFF
--- a/addon/-private/column-tree.js
+++ b/addon/-private/column-tree.js
@@ -589,6 +589,7 @@ export default EmberObject.extend({
     this._super(...arguments);
 
     this.token = new Token();
+    this._root = null;
 
     this._sortColumnsByFixed = this.sortColumnsByFixed.bind(this);
     this._ensureWidthConstraint = this.ensureWidthConstraint.bind(this);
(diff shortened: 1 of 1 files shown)


addon/-private/column-tree.js
```diff
@@ -599,7 +600,10 @@ export default EmberObject.extend({
 
   destroy() {
     this.token.cancel();
-    get(this, 'root').destroy();
```
mixonic, line 602: Why was this destruction call insufficient?

mmun: If the `root` computed property is invalidated (i.e. `columns` changes) then old value of root is lost and never destroyed.

mmun: The new code in `destroy` isn't much different. It just avoids unnecessarily instantiating a root node if it wasn't already computed.

The main fix in this PR is to add the cleanup logic at the begining of the root CP computation.

mixonic (approved):

mixonic:
I see it now. Excellent.

PR #1018, opened 2023-08-22
Migrate to explicit logicalWidth values in test

mixonic:
Extracted from https://github.com/Addepar/ember-table/pull/948 to see if I can get a green build against `master`

As described at https://github.com/Addepar/ember-table/pull/948, this method to determine logical element width is more accurate than `offsetWidth`. `offsetWidth` can only be an integer.

### DIFF
--- a/addon-test-support/pages/ember-table.js
+++ b/addon-test-support/pages/ember-table.js
@@ -6,6 +6,17 @@ import EmberTableFooterPage from './-private/ember-table-footer';
 import EmberTableHeaderPage from './-private/ember-table-header';
 import EmberTableLoadingMorePage from './-private/ember-table-loading-more';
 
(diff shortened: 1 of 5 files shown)


addon-test-support/pages/ember-table.js
```diff
@@ -6,6 +6,17 @@ import EmberTableFooterPage from './-private/ember-table-footer';
 import EmberTableHeaderPage from './-private/ember-table-header';
 import EmberTableLoadingMorePage from './-private/ember-table-loading-more';
 
+function computedStyleInPixels(target, property) {
+  let stringValue = window.getComputedStyle(target)[property];
```
mmun, line 10: I searched around and confirmed this should always be in pixels.

mmun (approved):

PR #1054, opened 2023-12-15
Compare offsetHeight/getComputedStyle at scale

mixonic:
At very large heights based on many child elements, Chrome and even Firefox begin to diverge the computed height values from what you see at lower scales. This makes their comparison with offsetHeight less reliable at high scales.

However, at large height values the exactness of the height measurement also matters less. At the end, only a limited level of precision is desired in the scale value and later code throws away noise at the lower end.

So, use a more forgiving comparison methodology (% shift in the change) vs requiring the two values to be only a single digit off.

I've asserted this on a table in our product which triggers the exception with a height of 100877px, but not below that number. This fix as written avoids the exception.

### DIFF
--- a/addon/-private/utils/element.js
+++ b/addon/-private/utils/element.js
@@ -30,6 +30,10 @@ export function closest(el, selector) {
   return null;
 }
 
+function parseComputedStyleHeightToPixels(computedHeightStyleValue) {
+  return Number(computedHeightStyleValue.substring(0, computedHeightStyleValue.length - 2));
+}
+
 /*
  * Calculate the scale of difference between an element's logical height
  * (the offsetHeight or getComputedStyle height) compared to its rendered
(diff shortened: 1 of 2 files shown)


addon/-private/utils/element.js
```diff
@@ -61,18 +65,31 @@ export function getScale(element) {
...
 
   if (isNaN(computedHeightInPixels)) {
     computedHeightInPixels = offsetHeight;
-  } else if (Math.abs(computedHeightInPixels - offsetHeight) >= 1) {
-    throw new Error(
-      "EmberTable's getScale() utility can only work on elements where height as derived from getComputedStyle is reliable. Usually this means padding is present on the target element."
-    );
+  } else {
+    let [min, max] = [computedHeightInPixels, offsetHeight].sort();
+    let difference = max - min;
```
mmun, line 76: `Math.abs(computedHeightInPixels - offsetHeight)` is simpler

mmun (approved):

addon/-private/utils/element.js
mixonic: Yes but I need `max` below this anyway. 👍

PR #1108, opened 2024-07-09
types: add and use TableRowMeta interface

lukasnys:
## Why?
I'm using the `ember-table` package with TypeScript and noticed I got errors in my template when I attempted to use any of the `rowMeta` properties.

## What?
Define the `TableRowMeta` interface based on [addon/-private/collapse-tree.js](https://github.com/Addepar/ember-table/blob/master/addon/-private/collapse-tree.js) and use it instead of `unknown` in the type declarations.


kpfefferle (approved):

lukasnys:
@kpfefferle any chance a new minor release could be cut for the changes of this PR as well as for !1106 and !1105?

kpfefferle:
@lukasnys Released as `6.0.0-2`

PR #1117, opened 2024-07-31
fix(ts): fix types for ember-tr onClick and onDoubleClick

lukasnys:
While using the `@onClick` argument on a `EmberTr` component, I noticed the value I was getting back didn't match its type.

I based the type off of the implementation of the `click` and `doubleClick` function here: https://github.com/Addepar/ember-table/blob/393d36f22fce0c513133d1c165a54332490d4fcc/addon/components/ember-tr/component.js#L97-L131


kpfefferle (approved):

mixonic:

PR #1118, opened 2024-08-01
fix sticky-polyfill causing crash on destroy

lukasnys:
When a ember-table is initially rendered without a `tfoot`, and a `tfoot` is added after the initial render (e.g. due to a loading state), the table will cause a crash on destroy.

This happens because the `TABLE_POLYFILL_MAP` does not have an entry for the specific `tfoot` element and thus, we call `.destroy()` on `undefined`.


kpfefferle (approved):

mixonic:

PR #1157, opened 2025-01-08
[CI] Add embroider-safe scenario

bantic:
Follows the guidance at https://github.com/embroider-build/embroider/tree/main/packages/test-setup and adds `embroider-safe` and `embroider-optimized` scenarios to the ember-try config.

This PR only enables the embroider-safe scenario because the embroider-optimized one has some failures, largely due to the fact that `ember-cli-addon-docs` is not installed by default (it is only installed for the ember-try scenario named `ember-default-docs`) but it is referenced from the dummy app in multiple places. The embroider-safe CI run succeeds, but the embroider-optimized (rightly, imo) complains about it.

The embroider-optimized scenario also fails because some of the tests use dynamic components (here, in specific: https://github.com/Addepar/ember-table/blob/04a4e5e74400deac7f9c1b4a4e5267cfa1760ed9/tests/helpers/generate-table.js#L64), and this could likely be fixed using some of the techniques described at https://github.com/embroider-build/embroider/blob/main/docs/replacing-component-helper.md.


.github/workflows/ci.yml
```diff
@@ -41,9 +41,12 @@ jobs:
     timeout-minutes: 10
 
     strategy:
+      fail-fast: false
```
bantic, line 44: Semi unrelated. This allows all ember-try scenarios to run to completion.

package.json
```diff
@@ -100,6 +103,7 @@
   },
   "homepage": "https://Addepar.github.io/ember-table",
   "resolutions": {
+    "@ember/test-waiters": "^3.0.2",
```
bantic, line 106: ember/test-waiters at 3.0.1 is not compatible with embroider. We fix this using a resolution because it's a secondary dependency of ember-a11y-testing, and bumping ember-a11y-testing to the ^5.2.1 range did not pick up the newer ember/test-waiters.
In a follow-up I'll attempt to upgrade ember-a11y-testing, at which point we can probably remove this.

package.json
```diff
@@ -51,12 +51,15 @@
     "@addepar/sass-lint-config": "^2.0.1",
     "@addepar/style-toolbox": "~0.8.1",
     "@ember/optional-features": "^2.0.0",
+    "@ember/string": "^3.0.0",
     "@ember/test-helpers": "^2.8.1",
+    "@embroider/test-setup": "^4.0.0",
+    "@glimmer/component": "^1.1.2",
```
bantic, line 57: embroider complains if this isn't present because the dummy app imports from it here: https://github.com/Addepar/ember-table/blob/3bb089665b2031e74c86bf9b5691976f9249ef18/tests/dummy/app/components/examples/aborting-a-selection/component.js#L1

mixonic (approved):

mixonic:

PR #1171, opened 2025-04-14
Set/Remove `isResizing` property on `TableColumnMeta`

vishaalagartha:
## Summary
Adds `isResizing` property to the columnMeta.

## Description
Consuming applications need the ability to style table columns based on whether or not they are being resized. Currently, ember table only add a `is-resizing` class to the container, but no modifications are made to the column.
By adding the `isResizing` property, consumers can style appropriately. For example, using the something akin to the snippet below in a consuming application:
```
 <EmberTd @class={{if r.columnMeta.isResizing "is-resizing")}}>
```

Note the `is-resizing` class being added/removed appropriately on the `td` element.

### DIFF
--- a/addon-test-support/pages/-private/ember-table-body.js
+++ b/addon-test-support/pages/-private/ember-table-body.js
@@ -21,6 +21,7 @@ export const BodyCell = PageObject.extend({
   isFirstColumn: hasClass('is-first-column'),
   isLastColumn: hasClass('is-last-column'),
   isSlack: hasClass('is-slack'),
+  isResizing: hasClass('is-resizing'),
 });
 
 export default PageObject.extend({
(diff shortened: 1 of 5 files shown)


bantic (commented):
This looks ok to me.

The biggest concern would be that this change introduces a performance regression. Can you add any detail on why it wouldn't have a negative perf impact? I don't see any new objects created here, which is good. Are we already modifying properties on the `node` object during a resize? That would be an indicator that this is safe, perf-wise. I'm wondering if it's possible that there are a lot of subscribed listeners to the `node` and by setting that property on it there would be a cascade of additional work -- but I don't think that'd be the case.
In a very large table (1000s+ rows), does that equate to 1000s of these `node`s, or do they only represent the virtualized visible rows?
Considering that E-T already very clearly does work during a resize (the `_updateResize` method that recalculates column widths), it would be surprising if the changes in this PR had any additional perf overhead. But I think it's worth just clearly thinking it through first.

Only blocking this is, could you add a test to this PR that will ensure the new behavior doesn't regress later on?

bantic (approved):
lgtm! Thanks for your patience and iteration!

Issue #995, opened 2023-08-03
Proposal: Ember Table 6


mixonic:
Its that time again! Building on https://github.com/html-next/vertical-collection/issues/400, I propose a new major version of Ember Table.

* Support Ember 3.28+
* Require vertical-collection v5 https://github.com/html-next/vertical-collection/issues/400
* Refactor Ember Components to Glimmer Components https://github.com/Addepar/ember-table/tree/master/addon/components (if apps inherit and extend from these, that may be a challenge and make this not worth doing)
* Change various calculations which rely on browser APIs that were presumed to return accurate values, but actually return rounded integers: https://github.com/Addepar/ember-table/pull/948. This improved scaling calculations an stability & accuracy in positioned items.

What else?

gwak:
Hey that's great news! Maybe we could consider
- adding the possibility of providing a custom component for the checkbox instead of relying css only to style it to correspond to one's design system ?
- floating table header without the need to specify a fixed table height, it's doable in user land but cumbersome (we do it with 2 ember-tables synchronized in a parent component)

apellerano-pw:
The most recent project I built with ember-table had a highly stylized table and we had a lot of issues getting table borders to work with `position:sticky` table elements.

You can see these visual glitches even on [ember-table's own documentation](https://opensource.addepar.com/ember-table/docs/guides/main/table-customization#cell-templates). The header specifies a border style, and it allows content to "peek thru" where the border would be as you scroll. To save you a bunch of SO dumpster diving, this isn't ember-table's fault. It's due to bugs in Chrome's table renderer that are in some cases a decade old, and blocked by undefined behavior in the HTML table spec. Help isn't coming.

One possible solution is to avoid `position:sticky` altogether in tables. You may have found yourself in this situation before and solved it by creating 2 or more ember-tables and keeping their scroll position in sync somehow. E.g. your header is an ember-table with 1 row. your fixed left column is an ember-table with 1 column. and finally the rest of the content is a third ember-table.

This is a heavy handed workaround, and if it's bothersome enough, perhaps ember-table could implement sticky rows & cols by moving the "many tables, synchronized" into the lib as an implementation detail; saving all the ember-table consumers. And ditch `position:sticky` and all its foreverbugs

apellerano-pw:
Thought of another problem with the use of `position:sticky`. The resize handles can't overlap the line between 2 header cells, which is what users expect.

Expectation: you grab the line to resize

[...]

PR #1172, opened 2025-05-27
Updated eslint and test dependencies

ijlee2:
## Background

We use and maintain a fork of `ember-table` at work, and noticed that the pod structure is still used for components. Generally speaking, [`ember-source@6.0.0` removed pod support for components](https://deprecations.emberjs.com/id/component-template-resolving), so `ember-table` may end up becoming a blocker to developers who want to update Ember to v6.

I suspect that the `ember-release` scenario (i.e. `6.4.0`) is currently passing, because the classic components make use of the `layout` key. Removing the pod structure now would help maintainers and contributors glimmerize these components (Glimmer + pod doesn't work) and introduce the v2 addon format.

## What changed?

In this pull request, I updated `eslint` and test dependencies so that we can keep this project up-to-date. The actual unpodding is done in #1173.


.gitignore
```diff
@@ -9,16 +9,12 @@
...
 /libpeerconnection.log
 npm-debug.log*
 yarn-error.log
 testem.log
-
-# ember-try
-.node_modules.ember-try/
-bower.json.ember-try
-package.json.ember-try
-yarn.lock.ember-try
```
ijlee2, line 24: Not needed in `ember-try@v4`.

mixonic (approved):
Approving to allow tests to advance

config/environment.js
```diff
@@ -1,19 +0,0 @@
...
-    environment,
-  };
-
-  if (environment === 'development') {
-    ENV['ember-a11y-testing'] = {
-      componentOptions: {
-        turnAuditOff: true, // Change to true to disable the audit in development
-        visualNoiseLevel: 2,
-        axeViolationClassNames: ['alert-box', 'alert-box--a11y'],
-      },
```
mixonic, line 14: Where does this configuration migrate to?

mixonic:
Seems like Ember 3.12 tests fail reliably here- But good news, we can just drop Ember 3.12 from the CI suite on `master` per https://github.com/Addepar/ember-table/issues/995. Just needs a PR to do it. I can follow through on that and rebase this again (I rebased it once to address a merge conflict)

mixonic:
I've fore pushed to rebase atop `master`. This should remove Ember < 3.28 from the CI system.

config/environment.js
ijlee2: The configuration was meant for the dummy app, so I'd guess `tests/dummy/config`. You'll want to check if `ember-a11y-testing` still support these options.

https://github.com/ember-a11y/ember-a11y-testing/tree/v7.1.2-ember-a11y-testing

PR #1183, opened 2025-07-03
Drop < Ember 3.28 from CI

mixonic:
In accordance with https://github.com/Addepar/ember-table/issues/995, drop versions of Ember older than 3.28 from the test suite.


mixonic:
To unblock https://github.com/Addepar/ember-table/pull/1172

PR #1190, opened 2025-08-19
Mark rows as even or odd for styling purposes

kpfefferle:
- Addepar's new visual look includes zebra-striping tables
- Since ember-table uses vertical-collection to occlude rows during scroll, this can't be reliably done with CSS selectors like `:nth-child(even)` or `:nth-child(odd)`
- A number of Addepar teams have manually added `class="{{if (eq (modulus-of body.rowMeta.index 2) 0) 'is-even' 'is-odd'}}"` to their ember-table row component calls to add stable row class names that indicate which rows have an even index vs an odd index
- This PR brings this capability into ember-table itself so that we (and other consumers) don't have to repeat this pattern over and over in order to give ember-tables a zebra stripe styling

### DIFF
--- a/addon/components/ember-tr/component.js
+++ b/addon/components/ember-tr/component.js
@@ -48,7 +48,7 @@ export default Component.extend({
   layout,
   tagName: 'tr',
   classNames: ['et-tr'],
-  classNameBindings: ['isSelected', 'isGroupSelected', 'isSelectable'],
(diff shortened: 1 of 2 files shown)


kpfefferle:
This same test failure seems to also be happening on `master`, so I don't believe it has anything to do with this PR 🤔

DanMonroe (approved):
Nice, thanks Kevin.  This will clean up some code for us.

PR #1252, opened 2026-09-14
Update ember-raf-scheduler

eliasdawson-addepar:
Update dependency `ember-raf-scheduler` and remove usage of `registerWaiter` method. `registerWaiter` is deprecated and a no-op as of `ember-raf-scheduler@0.5.0`. https://github.com/html-next/ember-raf-scheduler/blob/master/src/register-waiter.js.

Updating `ember-raf-scheduler` required additional changes. `@html-next/vertical-collection@v4` depends on `ember-raf-scheduler@0.3.0`. The version conflict resulted in tests in the `embroiderSafe` scenario using `0.3.0` and thus not waiting properly. This PR updates `@html-next/vertical-collection` to v5 to prevent this conflict.

Updating `@html-next/vertical-collection` to v5 also required additional changes to the template and build config.


.github/workflows/ci.yml
eliasdawson-addepar, line 1: Unrelated to these changes, but StepSecurity failed the build because these were unpinned. @arshabh-agarwal not this is also blocking all Dependabot and other automated PRs. If we don't land this PR for some reason, we should pull this out to unblock.

addon/components/ember-tbody/template.hbs
eliasdawson-addepar, line 1: Please check this carefully. I'm not very familiar with this format, and the `{{#vertical-collection...}}...{{else}` construction in particular.

mixonic (approved):
Wow very cool!

